PR #20, opened 2024-05-25
Earbleed Redux

NPC1314:
## About The Pull Request

Normalizing sounds to organized levels is a good thing.
This handles footstep, ambient loops, ambient music and instruments and ending credits.
Also adds alternative ambient music for tavern (1 is from Blackstone, 3 new) and bathhouse, and some guitar songs for another thing.

## Why It's Good For The Game

Overall its going to be more pleasant playing, will need a follow up though.
Its good enough for testing, need more ears. Not perfect yet but a big step forward.
Can add back the old 3 tracks put on ice easy enough but the new tracks deserves to get a shot.

PR #22, opened 2024-05-26
Gato basado (re-opened)

hermaplusplus:
> ## About The Pull Request
> 
> Inn cat gets some new behaviours
> * hunts rats, sort of. Only when walking around and only those close, but still.
> * walks around a little bit more often
> * moves over when someone steps in same tile, doesnt block movement any more
> * doesnt like being petted by vampires and dark elves, because why not have a quirk. Hisss
> 
> ## Why It's Good For The Game
> 
> Inn cat is terribly boring currently

Originally from #19 (reverted in #20 due to merge without review).

> ## About The Pull Request
> 
> Added the relevant ishelpers from Blackstone to the right place, less sloppy
> Tweaked the rat eating snippet to prevent double dipping rats
> Threw in a fix for the Collapse abuse, basically is a knockdown self thing now instead of instasleep

Addition from #23.

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/cat.dm
+++ b/code/modules/mob/living/simple_animal/friendly/cat.dm
@@ -1,5 +1,3 @@
-#define isdarkelf(A) (is_species(A, /datum/species/elf/dark)) // guess we don't have ishelpers yet
-
 //Cat
 /mob/living/simple_animal/pet/cat
 	name = "Inn cat"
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/simple_animal/friendly/cat.dm
```diff
@@ -295,3 +314,29 @@
...
+	. = ..()
+	if((isdarkelf(M)))  // l´cursed bonbonbon
+		visible_message("<span class='notice'>The cat hisses at [M] and recoils in disgust.</span>")
+		playsound(get_turf(src), 'sound/neu/cathiss.ogg', 80, TRUE, -1)
+		dir = pick(GLOB.cardinals)
+		step(src, dir)
+	if(M.mind && M.mind.has_antag_datum(/datum/antagonist/vampirelord))
+		visible_message("<span class='notice'>The cat hisses at [M] and recoils in disgust.</span>")
+		playsound(get_turf(src), 'sound/neu/cathiss.ogg', 80, TRUE, -1)
+		dir = pick(GLOB.cardinals)
```
HaultyAnonie, line 339: Wouldn't it have a chance to also recoil towards the darkelf / vampirelord if you just pick random cardinals?

hermaplusplus: @NPC1314

NPC1314:
I'll reupload from my fork with a little upgrade based on the feedback, ze cat will return

PR #31, opened 2024-05-27
quickfixers....

Dorodomki:
## About The Pull Request

quickly fixes the general broken things of the map, doesn't add new things but replaces the shelfs with normal tables until they are fixed, currently they are completely broken and can't be crafted, moved or destroyed, fixes some invisible tables on the inn, adds new open doors on the towner departments, now they can close their doors from inside for privacy and go AFK without get stolen

## Why It's Good For The Game

makes the life a bit more happy after a few weeks of content pause

## Pre-Merge Checklist
didn't checked anything but all should be good!

images and details of the work done

-image

-image


YuiY1997:
hey so I also have a mapping PR up and they'll override each other. We should both make sure we keep a copy of our changes to make sure that they aren't lost if one of our PRs gets overridden by the other

Dorodomki:
> hey so I also have a mapping PR up and they'll override each other. We should both make sure we keep a copy of our changes to make sure that they aren't lost if one of our PRs gets overridden by the other

you are right homie

PR #36, opened 2024-05-27
creecher drop fix

cornercube:
## About The Pull Request
fixes an oversight in animal droppings

volfs now drop hide and fur (confirmed to be their intended behaviour)
bugspiders drop honey (they were also supposed to do this)
rous drop fur (it just makes sense)

adds a greater mole with improved drops and higher stats than the lesser mole (no spawn yet)
drops lots of hide and meat
also adds descriptions to the moles, just for fun

## Why It's Good For The Game
bugfix
rous dropping fur makes more sense
annoyed by the fact that lesser moles exist without reference to other moles

## Pre-Merge Checklist


cornercube:
> A greater version of a mob really needs an alternative sprite, even if it's just a slight colour change. It should be recognisably different from the normal one.

I will just keep that commented out until I can get a sprite for it, in case someone is interested
just so people can get better volf drops quicker

PR #48, opened 2024-05-28
opens up Court Magician to Tieflings and Dark Elves

Valkrae:
## About The Pull Request
adds dark elf and tiefling to the race selection for Court Magician

## Why It's Good For The Game
Tieflings and Dark Elves should be able to be Court Magicians, as they are able to be Merchants. The role functions as a mostly RP-based role as it is, anyway, so I don't see any negatives to opening up the racial selection for the role.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/magician.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/magician.dm
@@ -8,6 +8,8 @@
 
 	allowed_races = list("Humen",
 	"Humen",
+	"Dark Elf",
+	"Tiefling",
 	"Elf",
 	"Aasimar",
 	"Half-Elf")


BadAtThisGame302:
I personally don't think Court Mage and Merchant are the same, Court Mage is akin to Steward, Hand or Captain due to being one of the Kings main advisors which none can be Drow or Tiefling if I am not mistaken.

Valkrae:
> I personally don't think Court Mage and Merchant are the same, Court Mage is akin to Steward, Hand or Captain due to being one of the Kings main advisors which none can be Drow or Tiefling if I am not mistaken.

That comparison was meant more to describe how they arrived to Stonekeep, rather than their role in the town. I've always seen the Merchant as an external addition from the Guild, rather than someone who has lived in the Town for a while. I feel like this logic also applies to the Court Magician, in that they are likely not from Stonekeep as well.

As for the Steward/Hand/Captain comparison, I disagree on that.

Valkrae:
other comments by machaa on other things have made this obsolete so goodbye drow court magician 😔

PR #88, opened 2024-05-30
updates graveyard and market rework for new map changes

YuiY1997:
fixes conflicts

## About The Pull Request

Adds markets and updates graveyard in the southeast quarter of the city. Edits sewers to fit new crypt. Updates townhouse and gives it a second floor.

## Why It's Good For The Game

More room to roleplay, makes it so the graveyard can be more private, allows burial for important people in the crypt instead of the mud further emphasizing class division, gives people markets. A lot of people like setting up their own food stands and such and now have room to do this.

## Pre-Merge Checklist

image
image
image

### DIFF
--- a/_maps/map_files/roguetown/roguetown.dmm
+++ b/_maps/map_files/roguetown/roguetown.dmm
@@ -12627,10 +12627,6 @@
 	icon_state = "plating2"
 	},
 /area/rogue/outdoors/town/roofs)
-"tHZ" = (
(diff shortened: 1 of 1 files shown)


YuiY1997:
alright applied some fixes after the testmerge. Should be 100% good to go now.

YuiY1997:
okay so the problem where turfs under the stairs. Both are totally dealt with now. Should work fine now. Thanks to everyone who reported problems while testing

PR #104, opened 2024-06-01
townhouses and homeless camp

YuiY1997:
## About The Pull Request
replaces the unlit, unused lot in the southeast corner of the map with a row of townhouses and a homeless camp for beggars.

## Why It's Good For The Game

More places to roleplay, more houses, gives beggars a "home base"

image
image
image

## Pre-Merge Checklist


X404ErrorX:
i dont know if this is posible but can you add keys for the houses? they can spawn inside the houses or next to the doors so people can grab them and claim the house.

YuiY1997:
It is doable to add keys, but I don't want to add a bunch of keys to the steward's as well. Instead the doors are able to be deadbolted from the inside.

X404ErrorX:
I don't think it's a problem that people get locked in since only those who set ready will spawn in the houses so they can grab the key quickly in the start of the round, the late join towners spawn in the north area of the bath houses on the dock. 
I understand that the issue of keys can be somewhat problematic, but I think it loses the point of having a base/house if you can't even close it to prevent people from stealing your things and so on, you have to keep in mind that when you leave the house It will always remain open since there is no way to close or open it from the outside. Although on the other hand perhaps we could add a lock and key crafting system to solve this.

Optimism333:
It would really be nice if the houses had keys. Having a key to the house gives you a mechanical sense of ownership, and more attachment than a bolt. Because you can store things in the house and be assured that they'll at least have the security of a locked door.

YuiY1997:
I will see how this one goes after it's merged and add keys when I map in the building for surgeon.  I will have to make new keys for surgeon anyway, so I'll just code in keys for every house too.

PR #105, opened 2024-06-01
Farm Revamp

k0n1n:
## About The Pull Request

Braziers added near the palisades **_(No more need to leave the farm to look for stones and wood to make a hearth between the soil)_**
No more keys in Soilson's inventory that don't open the farm doors **_(Bug "fixed")_**
Five keys added to the shelf **_(No more fighting over who gets the key and becomes the "gatekeeper" instead of taking care of the crops)_**
Five scythes added to the table **_(You can take the scythe now with you when going to the town instead of leaving it on the farm and having someone make it disappear)_**
Five bags added to the table outside **_(Initial bags to speed up the storage of crops inside the chests)_**
Tree removed to the northwest of the palisade and a palisade placed to secure it
Two wall sconces added on the second floor **_(No more walking in the dark to find the bed)_**
A brazier added near the well **_(No more fetching water in the dark)_**

## Why It's Good For The Game

It's all very time-consuming when there's not more than one soilson to operate on the farm, making the life of one or all soilsons easier, making the farm work faster and more productive.

## Pre-Merge Checklist


NPC1314 (changes requested):
Looks like a good bugfix, if branch updated the farm changes pasted over to the up to date map, should get prio for map PRs in my view.

Minor annoyance is the failure to document sewing skill drop from 2 to 1 but it seems unimportant

Optimism333:
Why drop sewing skill and the intelligence bonus by one without saying anything, though? Just remove those changes

PR #110, opened 2024-06-02
Redistributes some stats and equipment for the rogue class

1N7Slayer:
## About The Pull Request

The primary objective of this PR is to make rogues less irritating to deal with due to the possession of a crossbow. In exchange, they get a bow and some arrows and archery skill instead of their crossbow skill. 

I also made them a bit squishier in the CON and END department while giving them an extra +1 to INT for skill check probability. In certain situations, such as for elven thieves, they might make mildly better crafters and skill learners. In the future, I may take a better look at all class distribution lineups for stats.

## Why It's Good For The Game

It doesn't make much sense for a thief to have a crossbow in a game and roleplay sense, it is a really awful self defense weapon and is more geared for assassinating one particular person and running away. This is awful for the skirmishing playstyle expected of a rogue. Instead, I gave them a bow and quiver of arrows, so they can slow down attackers and then still get away with it, but without the absurd upfront damage afforded by a crossbow. Because arrows do not do very much damage, rogues need to shoot a lot of them to really endanger someone. Legshots still are viable in this instance, but can pin more than one guard at a time to allow their escape -- at the cost of the arrows, not easily replaced for a criminal. End of analysis.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rogue.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rogue.dm
@@ -1,6 +1,6 @@
 /datum/advclass/combat/rogue
 	name = "Rogue"
-	tutorial = "Where is the gold?"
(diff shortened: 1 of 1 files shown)


iskanderone:
Closing this for now, actually, thinking about something considerably more comprehensive than this.

Issue #134, opened 2024-06-05
[Bug]: Restart vote issues


NPC1314:
### Checks

- [X] I have searched the bug with a few keywords, and I confirm this bug was not yet reported.

### Location

RogueTown

### Describe the issue

Restart vote results not implemented correctly all the time. Lack information on if its by type, so need more feedback.
Do player-initiated restart votes work?
Do admin-initiated restart votes work?
Do automated restart votes work?
If you can clarify any of the above three questions please drop a line.

PR #135, opened 2024-06-05
Quick n dirty

NPC1314:
## About The Pull Request
Re:  #134 
Should return us to the old bug instead of the new, but needs to be babysat on testmerge a little, can't guarantee it works well
## Pre-Merge Checklist


ThePainkiller:
I believe gating the votes appearing behind a `if(roundvoteend)` would also solve these appearing AFTER a roundend vote passed, so that it doesn't conflict between themselves and prolong the round

PR #137, opened 2024-06-05
Captain's Shite

Max-023:
## About The Pull Request
Gives the Captain a proper helmet, plate leggings, and gloves for protection to fit their stature as a Knight. Also adds the 'Ser' honorific so there is no confusion as to their position.

## Why It's Good For The Game
Because not only were Officers Knights, why the fuck wouldn't the standing head of the Garrison and his high salary not have a helmet and gloves?

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/sheriff.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/sheriff.dm
@@ -19,11 +19,21 @@
 	give_bank_account = 120
 	min_pq = -4
 
+/datum/job/roguetown/manorguard/after_spawn(mob/living/L, mob/M, latejoin = TRUE)
+	..()
+	if(ishuman(L))
+		var/mob/living/carbon/human/H = L
+		var/prev_real_name = H.real_name
+		var/prev_name = H.name
+		var/honorary = "Ser"
+		H.real_name = "[honorary] [prev_real_name]"
+		H.name = "[honorary] [prev_name]"
+
 /datum/outfit/job/roguetown/captain/pre_equip(mob/living/carbon/human/H)
 	..()
 	head = /obj/item/clothing/head/roguetown/helmet/heavy/knight
 	gloves = /obj/item/clothing/gloves/roguetown/plate
-	pants = /obj/item/clothing/under/roguetown/trou/leather
+	pants = /obj/item/clothing/under/roguetown/platelegs
 	armor = /obj/item/clothing/suit/roguetown/armor/brigandine/captain
 	neck = /obj/item/clothing/neck/roguetown/gorget
 	shirt = /obj/item/clothing/suit/roguetown/shirt/shortshirt


code/modules/jobs/job_types/roguetown/nobility/sheriff.dm
```diff
@@ -19,9 +19,21 @@
...
 
+/datum/job/roguetown/manorguard/after_spawn(mob/living/L, mob/M, latejoin = TRUE)
+	..()
+	if(ishuman(L))
+		var/mob/living/carbon/human/H = L
+		var/prev_real_name = H.real_name
+		var/prev_name = H.name
+		var/honorary = "Ser"
+		H.real_name = "[honorary] [prev_real_name]"
+		H.name = "[honorary] [prev_name]"
```
BadAtThisGame302, line 30: Remove honorary Ser and make it Captain (gender neutral), or add a female equivilant (Dame)

Max-023: Not to 'erm actually' but Dame isn't a female equivalent in most European Knight orders, as Dame was used for women coinciding with the traditional views of them unrelated to chivalry or combat at all. Eventually women were even given the title of 'Sir' when knighted by the British throne.

Personally, I believe that Ser works as a catch all and those not officially part of the court's forces can use the Dame title if they want to say it's part of a different Knightly order. Not to mention it wouldn't be fitting to call them 'Captain' over 'Ser', as you are only supposed to call those of lower rank (which would later be called NCOs in modern militaries) by their rank like 'Sergeant' to denote respect of veterancy. For nobles, you always call them ser or (if they are your liege lord) lord.

BadAtThisGame302: image
image
image

Max-023: Those are all modern examples. Historically, it was rare for any woman to take up arms at all at all and most rare examples outside the English were given the title of 'Sir' as the gender neutral title. More modern orders founded after WWI, and those that somehow survived from prior, would change this as women sought to enter the orders with the boom of feminism in western societies.

As for the English, not only is 'Ser' the gender neutral title for those of martial class, but even in Knightly orders the female specific changes depending on the context, and all women specific were not related to chivalry. The Order of the Garter, for instance, called them (and still do) 'Lady Companion' rather than 'Royal Knight' to show that they are just companions. This is because in England, the title was more of a title for nobility as a means of prestige, while historically in mainland Europe it was to denote either a) the individual was an Officer, b) they were a Chevalier (heavy cavalryman), or C) or part of a Monastic order. All of which required the individual to be a nobleman already and noble women were mostly barred from service in retinues instead to be married off.

If we are to simulate medieval societies that they were suddenly okay with women having equal standing and opportunities of men in their class without discrimination, the majority would still only receive the title 'Sir' or an equivalent because Dame is, debatably, sourced from the French word 'Madame' and most English knightly orders came about in England when French influence was at its highest. So for realms with gendered languages, like German and French, the masculine term would also double as the gender neutral term. Thus, Ser would be applicable.

Regardless, I've spoken to a better coder than myself and they want to make it a toggle the player can select. So for those playing female captains, they can choose either 'Dame' or 'Ser' rather than have an anachronistic title of 'Dame' forced onto them.

ThePainkiller: Will you be pushing a code snippet about the prompt for the honorary to apply, or will you do an if check for gender and leave it as Ser/Dame?
Also since you're going about adding a female version of the title, why must the male one go neutral and not "Sir", as it should?

Max-023: How the toggle will work is better to ask the individual who said they would after this PR is merged. So for now, the gender neutral 'Ser' will just apply to all.

As for once the change is in, yes of course the male chars will also have a choice between the gender neutral 'Ser' and masculine 'sir' as to keep parity between the two genders.

messagecypher:
the plate chausses sprite is still broken, recommend you use chainmail chausses in the meanwhile

PR #148, opened 2024-06-08
Lobby Music Normalization

NPC1314:
## About The Pull Request

Overhauling the entire audio landscape in one go is too unwieldy, and if I can escape the insane mixing in the lobby while testing this would save my sanity. Thus this, just adjusting some lobby music and a few music loops that plays intermittently (ambience etc still unfixed) The extreme outliers pruned down to -20 lufs, I find it much less unpleasant to hear, if anything might be too conservative but baby steps is still in the right direction.

## Pre-Merge Checklist


NPC1314:
For some reason it ended up too quiet IMO, going to tweak it up a lil

PR #150, opened 2024-06-08
Downtunes whips to no longer be able to delimb, reduces their combat effectiveness

1N7Slayer:
## About The Pull Request

Whips as it stand have ridiculous AP on crack (aka, the only combat intent worth using), delimb capability and can remove heads in 7 hits, as well as stun on hit with high enough strength due to the weapon having the BLUNT feature on the attack type, in addition to having up to 3 range.

Their stamina costs mean nothing because they are not heavy weapons. They are blatantly overtuned, and I have reduced their combat stats accordingly, especially on crack. You can make a whip off a single volf you kill in the woods. Imagine if a simple spear had 3 range chop and knockdown but with no stamina cost. That would be ridiculous.

## Why It's Good For The Game
Whoever designed the whip without playtesting it was on crack. Overtuned weapons are bad for the game, mmkay?

## Pre-Merge Checklist


ThePainkiller:
There's an issue to this, multiple I'd say:

1. You changed the weapon class to `BCLASS_STAB` rather than `BLUNT`, which means whips can cause gruesome cuts considering they have 19 force at minimum
2. This also makes whips able to be used to carve messages in walls
3. This causes whips to cause bleeding, the main reason they used `BLUNT` was because they cause hematomas, their intended type of wound, but there's the issue of blunt attacks penetrating armor A LOT since they're used for maces

My suggestion is to create a new blunt-type `BCLASS` exclusively for whips, and their own intent datum, that causes hematomas but has absolutely zero armor penetration, and no delimb chance.

1N7Slayer:
smart one sec

code/game/objects/items/rogueweapons/melee/whips.dm
```diff
@@ -49,11 +49,11 @@
 
 /datum/intent/whip/crack
 	name = "crack"
-	blade_class = BCLASS_CUT
+	blade_class = BCLASS_LASHING
```
ThePainkiller, line 52: Excellent! Two more suggestions:

This intent should have `canparry = FALSE` as you can't really parry with a whip in hand. It already has candodge as TRUE by default, so no need to change that one.
This intent should have a `chargetime` of at least 5, which is similar to how Flails work, to simulate the feeling of "winding up" the whip to crack with it.
Considering its dimished damage capabilities (and the fact it can't break bones anymore), you could make it have a `recovery` value of 5 so that it makes sense for it not to be as hard to strike again with it as a mace made of an iron weight on its end.

code/game/objects/items/rogueweapons/melee/whips.dm
```diff
@@ -15,6 +15,7 @@
 	parrysound = list('sound/combat/parry/parrygen.ogg')
 	swingsound = WHIPWOOSH
 	throwforce = 5
+	canparry = FALSE
```
ThePainkiller, line 18: `canparry` goes on the lash/crack intent datums, not the item....

SELFHELL:
While you are doing this, you should reduce the str requirement for the dungeoneer whip. I overdid it and forgor to fix

ThePainkiller (approved):

ThePainkiller (approved):

PR #152, opened 2024-06-09
Adds Delay to Pickpocketing

maaacha:
Adds a two second period where both players must be standing still for a pickpocket to take place. Time is likely to change depending on tests. No progress bar or message  appears to the pickpocketed while this is in progress so it's still stealthy, now you just can't instantly pick someone's pocket.

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -388,7 +388,7 @@
 						if(!(zone_selected in stealablezones))
 							to_chat(src, "<span class='warning'>What am I going to steal from there?</span>")
 							return
-						if(do_after(U, 6 SECONDS, target = V, progress = 0))
+						if(do_after(U, 2 SECONDS, target = V, progress = 0))
 							switch(U.zone_selected)
 								if("chest")
 									if (V.get_item_by_slot(SLOT_BACK_L))


code/_onclick/other_mobs.dm
```diff
@@ -388,31 +388,34 @@
...
+								V.dropItemToGround(picked)
+								put_in_active_hand(picked)
+								to_chat(src, "<span class='green'>I stole [picked]!</span>")
+							else
+								to_chat(src, "<span class='warning'>I didn't find anything there. Perhaps I should look elsewhere.</span>")
 						else
-							to_chat(src, "<span class='warning'>I didn't find anything there. Perhaps I should look elsewhere.</span>")
-					if(stealroll <= 4)	
+							to_chat(src, "<span class='warning'>I fumbled it!")
+					if(stealroll <= 4)
```
ThePainkiller, line 418: **Opinion:** Considering the chance to steal is a Xd6 diceroll where X is our pickpocketing skill level, at 4 (Expert) and upwards it becomes impossible to fumble. The formula should be fine tuned so that there's at least a minimal chance unless you're a Legendary pickpocket, or even at Legendary pickpocket leave it at a maximum of 10% chance to fail IMHO

1N7Slayer:
This is good, but might I suggest pickpocket speed scale based on skill? Experts should be able to do it fairly quickly, 2 seconds on the dot, and masters 1 second, legendaries should be able to swipe and leave.

PR #159, opened 2024-06-10
Assassino PR

77seven77seven:
## About The Pull Request

split up into seperate PRs because i wanna do something with my life also i can make myself feel good with two things merged :))))

made 4 macha with love

## Why It's Good For The Game

more antags based on stealth rather than fragging mostly

## Pre-Merge Checklist

### DIFF
--- a/code/controllers/subsystem/rogue/role_class_handler/drifter_queue/drifter_waves/drifter_wave.dm
+++ b/code/controllers/subsystem/rogue/role_class_handler/drifter_queue/drifter_waves/drifter_wave.dm
@@ -8,7 +8,7 @@
 	// Title of the job related to the job subsystem thats being made/equipped towards for the wave
 	job_rank = "Drifter"
 	drifter_wave_categories = list(DTAG_FILLERS)
-	advclass_cat_rolls = list(CTAG_PILGRIM = 13, CTAG_ADVENTURER = 11)
+	advclass_cat_rolls = list(CTAG_PILGRIM = 26, CTAG_ADVENTURER = 22)
 	class_cat_plusboost_attempts = list(CTAG_PILGRIM = 3) // Here you go buddy, have a 3 on pilgrims
 
 	wave_delay_time = 2 MINUTES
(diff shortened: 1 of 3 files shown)


1N7Slayer:
I want this to be merged so I can play captain or royal guard and assist in indicating why this is a bad idea as it stands.

code/modules/projectiles/projectile/magic.dm
```diff
@@ -528,6 +528,23 @@
 			var/obj/item/I = target
 			I.throw_at(throw_target, 200, 4)
 
+/obj/projectile/magic/swap
```
ThePainkiller, line 531: My god this is going to be extremely cancerous to play around. Please don't.

casualspacestation14enjoyer: it has been nerfed

casualspacestation14enjoyer:
> I want this to be merged so I can play captain or royal guard and assist in indicating why this is a bad idea as it stands.

okay more fragging antags coming up!

casualspacestation14enjoyer:
merge conflicts  fixed

NPC1314:
Theres literally no description

code/modules/projectiles/projectile/magic.dm
ThePainkiller: I think my comment didn't go through so I'll be very specific

- You're giving a MAGIC SPELL to an assassin, in a world where magic is very limited and very rare (we have like 1 or 2 sorceror slots, and then there's only the court wizard)
- This magic spell still SWAPS targets with the person you're shooting it for, it's a literally get out of jail card for almost every situation, don't matter the range, or cooldown, it's an unblockable and annoying to play around mechanic that just turns him into "rogue but harder to catch if caught"
- Rather than give him a literal door opening spell (which makes this person literally unable to be jailed and therefore MUST be killed, or have both arms cut off with a permanent cloth to the mouth), it would be better to belay that until the locks PR gets merged so that we can introduce a lockpicking proficiency instead

casualspacestation14enjoyer:
merge conflict fixed

PR #163, opened 2024-06-12
Health Potion Rework

iloveyoulots:
## About The Pull Request

1. Fixes health potion not restoring lost blood properly.
2. Decreases metabolization rate of health potions so they no longer metabolize near-instantaneously.
3. Sets health potion healing rate lower, now that they no longer trigger the whole contents in one tick.

## Why It's Good For The Game

Health potions currently metabolize near-instantly, rendering them useless for most people. This change will make it so they no longer metabolize instantly, instead they will now heal you slowly over time. Furthermore, the blood restoration effect was bugged, so I took the time to fix that while reading. This should also have the effect of patching the exploit where a watered down health potion can be used to the same effectiveness as a major one - as it was only the case due to the absurdly high healing rate, which only needed to exist to compensate for the absurdly high metabolic rate.
## Pre-Merge Checklist

- [X] You tested this on a local server.
- [X] This code did not runtime during testing.
- [X] You documented all of your changes.

### DIFF
--- a/code/game/objects/items/rogueitems/keys.dm
+++ b/code/game/objects/items/rogueitems/keys.dm
@@ -24,8 +24,6 @@
 				lockhash = rand(100,999)
 			GLOB.lockhashes += lockhash
 			GLOB.lockids[lockid] = lockhash
-	else if(!lockhash)
-		qdel(src)
 
 /obj/item/roguekey/lord
 	name = "master key"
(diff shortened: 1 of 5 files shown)


code/modules/roguetown/roguejobs/alchemist/reagents.dm
```diff
@@ -5,17 +5,17 @@
 	color = "#ff0000"
 	taste_description = "red"
 	overdose_threshold = 0
-	metabolization_rate = 20 * REAGENTS_METABOLISM
+	metabolization_rate = 0.5 * REAGENTS_METABOLISM
 	alpha = 173
 
 /datum/reagent/medicine/healthpot/on_mob_life(mob/living/carbon/M)
-	if(M.bleed_rate < 0.3)
-		M.blood_volume = min(M.blood_volume+2, BLOOD_VOLUME_MAXIMUM)
```
Schmorgisborg, line 13: M -> C
for previous line too, building and testing how it affects wounds & various damages

ThePainkiller:
Values are fair, just solve the merge conflicts please.

PR #170, opened 2024-06-12
Blackstone Port - "Diagonal Bow Shots Work" (Fixes bugs we have here)

Stutternov:
## About The Pull Request

Basically does two main things. Firstly, it stops arrow duplication via making diagonal arc mode work. Secondly, makes it so if you are lying prone you still get hit by arrows, if you are not the direct target, if you can stand, use items, or move. (Aka stops arrows from whizzing over you; you can't just lay down to avoid being shot.)

Also - makes arrows check in person is dead. If dead, whiz over. Credit to the person who made it below, not my work.

https://github.com/Blackstone-SS13/BLACKSTONE/pull/1122

## Why It's Good For The Game

Basically fixes an error we've had that Blackstone has fixed + makes it so dodging actually matters instead of just laying down on the floor to avoid getting shot by indirect fire.

## Pre-Merge Checklist

- [X] You tested this on a local server.
- [X] This code did not runtime during testing.
- [X] You documented all of your changes.

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -568,7 +568,9 @@
 	else
 		var/mob/living/L = target
 		if(!direct_target)
-			if(!CHECK_BITFIELD(L.mobility_flags, MOBILITY_USE | MOBILITY_STAND | MOBILITY_MOVE) || !(L.stat == CONSCIOUS))		//If they're able to 1. stand or 2. use items or 3. move, AND they are not softcrit,  they are not stunned enough to dodge projectiles passing over.
+			//If they're able to 1. stand or 2. use items or 3. move, AND they are not softcrit,  they are not stunned enough to dodge projectiles passing over.
+			//If they're dead they shouldn't be getting hit by indirect fire
+			if((CHECK_BITFIELD(L.mobility_flags, MOBILITY_USE | MOBILITY_STAND | MOBILITY_MOVE) && L.stat == CONSCIOUS) || L.stat == DEAD)		
 				return FALSE
 	return TRUE
 
--- a/code/game/objects/items/rogueweapons/ranged/bows.dm
+++ b/code/game/objects/items/rogueweapons/ranged/bows.dm
@@ -36,7 +36,7 @@
 	return
 
 /obj/item/gun/ballistic/revolver/grenadelauncher/bow/dropped()
-	..()
+	. = ..()
 	if(chambered)
 		chambered = null
 		var/num_unloaded = 0
@@ -68,7 +68,7 @@
 			BB.damage = BB.damage
 			BB.embedchance = 100
 		BB.damage = BB.damage * (user.STAPER / 10)
-	..()
+	. = ..()
 
 /obj/item/gun/ballistic/revolver/grenadelauncher/bow/update_icon()
 	. = ..()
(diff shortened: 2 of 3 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -568,7 +568,9 @@
 	else
 		var/mob/living/L = target
 		if(!direct_target)
-			if(!CHECK_BITFIELD(L.mobility_flags, MOBILITY_USE | MOBILITY_STAND | MOBILITY_MOVE) || !(L.stat == CONSCIOUS))		//If they're able to 1. stand or 2. use items or 3. move, AND they are not softcrit,  they are not stunned enough to dodge projectiles passing over.
+			//If they're able to 1. stand or 2. use items or 3. move, AND they are not softcrit,  they are not stunned enough to dodge projectiles passing over.
+			//If they're dead they shouldn't be getting hit by indirect fire
+			if((CHECK_BITFIELD(L.mobility_flags, MOBILITY_USE | MOBILITY_STAND | MOBILITY_MOVE) && L.stat == CONSCIOUS) || L.stat == DEAD)
```
ThePainkiller, line 573: I'm very skeptical this work the way its intended to work, can you localtest and see if arrows do indeed pass over dead bodies?

ThePainkiller: Oh sorry, but most importantly, if arrows do NOT pass over unconscious people as well

Cre77:
dunno if it's this pr being testmerged, or some other thing - but right now the ARC shooting intent behaves exactly as the normal one. it will not fly over objects, people, and the arrow goes straight on the same z-level. tested in both the open-sky basin area and in the one-level bog

Stutternov:
> dunno if it's this pr being testmerged, or some other thing - but right now the ARC shooting intent behaves exactly as the normal one. it will not fly over objects, people, and the arrow goes straight on the same z-level. tested in both the open-sky basin area and in the one-level bog

I am not noticing this, it traveled fine into someone shooting over others on arc mode. Shown here.
image

Stutternov:
All that was changed directly here was the shooting mocde and arrow-dropping.

code/modules/projectiles/projectile.dm
Stutternov: Also, example of it not passing over a person who fell over but is not unconcious or dead. They take the arrow rather than it passing over top. 

image

ThePainkiller:
> All that was changed directly here was the shooting mode and arrow-dropping.

Mind when you can attaching screenshots of this branch and just this branch testing arc mode between Z-levels? People are saying this might be the cause of arc mode not travelling z-levels but I'm not sure, just want it clear so we can trim down where to hunt this bug for

Stutternov:
> > All that was changed directly here was the shooting mode and arrow-dropping.
> 
> Mind when you can attaching screenshots of this branch and just this branch testing arc mode between Z-levels? People are saying this might be the cause of arc mode not travelling z-levels but I'm not sure, just want it clear so we can trim down where to hunt this bug for

Tested it, it works between Z-levels. I shot from the tree north in that photo; it appeared to 'shoot' on your Z-level then drops down where you're aiming. No issue with it that I could replicate other than still spotting an arrow duping once or twice.

PR #176, opened 2024-06-13
Removes the physicker single slot lock

Plaguewalker:
## About The Pull Request
It removes the one slot limit for physicker making it more in line with the other pilgrim roles.

## Why It's Good For The Game
It brings it in line with the other pilgrim roles which aren't limited to a set number of slots with the exception of weaver. Once the player dies the role is already removed from the pool list of selectable classes for the player upon respawn for pilgrim. The role in itself puts fourth the person being that of a traveling doctor akin to how the cleric is being a traveling healer as well. It should not be limited to that of one person, I understand the limiter but this will allow a wider net currently catch any major bugs as well as prevent the role being squandered and lost to an afk player.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/weaver.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/weaver.dm
@@ -12,7 +12,6 @@
 	)
 	outfit = /datum/outfit/job/roguetown/adventurer/seamstress
 	category_tags = list(CTAG_PILGRIM, CTAG_TOWNER)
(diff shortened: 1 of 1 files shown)


ThePainkiller (approved):
To be honest, weaver should also have no slot restrictions. Slot restrictions should be for rare or certain combat classes.

Plaguewalker:
> To be honest, weaver should also have no slot restrictions. Slot restrictions should be for rare or certain combat classes.

Exactly, because those classes are actually rare and give you a bit of an edge like blade singer for example.

PR #188, opened 2024-06-15
WRESTLING CHANGES PORT from the good lads over at Blackstone

OctusGit:
## About The Pull Request

- https://github.com/Blackstone-SS13/BLACKSTONE/pull/542
Since this is a port im just gonna copy what's in the PR's description

As it is now, wrestling is near non-functional, with resist chances always being near-max, and punishes players with a chance to stun them for even trying to grapple. The purpose of this PR is to at least have it in a working state before adding any new wrestling-related features.

How it works:

- The difference in strength between attacker and defender modifies your resist chance by 10.

- The difference in wrestling skill reduces resist chance penalties from being aggro grabbed or prone. (more penalty types to come).

- Wrestling skill also determines the minimum chance for the defender to break out of an aggressive grab. The base is 70 for passive and 50 for aggressive. which is reduced by having a higher skill than the defender, but cannot be increased.

- This minimum chance prevents high strength chars from easily and always winning if they manage to grab you, but allows them to easily resist grabs themselves.

- Made changes to the struggling mechanic for attempting to upgrade a passive grab into a aggressive one. The chance to be stunned is much lower - unless the defender is also grabbing you as well, in which case it is the same as before.

- Failing to upgrade a grab also stuns the defender for half the duration. the overall stun duration is reduced.

## Why It's Good For The Game

Wrestling kinda sucks or doesn't feel as rewarding. So this is supposed to make it more viable and favorable for grapplers and fist fighters.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1086,7 +1086,7 @@
 	if(L.mind)
 		wrestling_diff -= (L.mind.get_skill_level(/datum/skill/combat/wrestling))
 
-	resist_chance += ((STASTR - L.STASTR) * 10)
+	resist_chance += ((STACON - L.STACON) * 10)
 
 	if(!(mobility_flags & MOBILITY_STAND))
 		resist_chance += -20 + min((wrestling_diff * 5), -20) //Can improve resist chance at high skill difference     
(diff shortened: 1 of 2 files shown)


CormanQuewye:
Strength already affects damage, knockback and knockdown; resisting out of a grab takes more stamina than applying it, so someone with high wrestling can grab and your choices are either resist or risk getting tackled and pinned, effectively putting you in a stamina drain loop.

tl;dr wrestling isn't bad and this just makes strength all the more absurd

Stutternov (approved):

PR #189, opened 2024-06-15
For Whom The Bell Tolls

ThePainkiller:
## About The Pull Request

- Bell ringing turned into a global sound, since it loses volume across distance and can't be heard through z-levels. Let the faithful know the church is calling. Let the faithless hear the sound of zeal wherever they are.

## Why It's Good For The Game

Makes the bell hearable through z-levels and the distance, unfortunately `orange()` acts for just one z-level and loses volume per distance travelled, so making it a global sound like announcements was the better idea.

## Pre-Merge Checklist


NPC1314 (approved):
Looks good.
Possible drawback is the bell being movable and the sound global, meaning some sweatgoblin can sneak around with it and ring the bell from weird places to call people to church, but not really a big one and can be handled if it becomes an issue

ThePainkiller:
> Possible drawback is the bell being movable and the sound global

The easy solution is to make the bell `anchored = TRUE`, unless you're talking about the bell being actually stealable from its resting place.

ThePainkiller:
Ohh I see, you think this is for the **PORTABLE** small bell. Nope, that one is still local. This affects the **BIG** church bell only, which is anchored and impossible to steal or move.

NPC1314 (approved):
Yeah, I thought it was the tiny one =)

Issue #180, opened 2024-06-13
[Bug]: ROUS Attack Each Other!


DAKKA-WAAAGH:
### Checks

- [X] I have searched the bug with a few keywords, and I confirm this bug was not yet reported.

### Location

RogueTown

### Describe the issue

While I'm testing my maps, I'm seeing the Rous attack each other in the terminal... no wonder I wasn't ever seeing the correct amount of Rous.

It would be more interesting if they banded together instead, and continued to spread if ignored.

### How to reproduce

Place two Rous in a room, they WILL kill each other.

### Round ID

N/A

### Staff ckey

N/A

NPC1314:
also /smallrats eats each other, needs blacklisting itself in food

ThePainkiller:
Closing as completed by #195

PR #195, opened 2024-06-15
Bugfix (180)

NPC1314:
## About The Pull Request

Fixes issue #180 
Technically the big rats attacking each other was according to design, but its better they have a chance to grow and become
more of a menace, a little too safe r/n
Also blacklists some foods for small rats (onions are toxic to rats and they can probably sniff out poison berries, also less
cannibalism means chance for more reproduction in dark corners of the sewers, good)

## Pre-Merge Checklist


ThePainkiller (approved):
I'd say it might be good practice to move this to where every global list INIT is, rather than have it on the file it's called. BYOND autism from initializing files in alphabetical order can really screw you over on this, plus it's neater. Otherwise looks fine.

PR #209, opened 2024-06-18
Marriage for All Couples [DONE]

InsightfulParasite:
## About The Pull Request
The gods will no longer curse you if your marrying someone of the same sex. Though now you will have to bite the apple first for your bride to take your surname.
**The above fact may be more important for nobles or the king**

I also added some more considerations for viable marriage candidates. Such as making sure they are alive and not SSD.

Notice: If two people at the wedding share the same real_name then the person being married will be selected randomly out of the two existing creatures. Dont know how much of a issue this is since im pretty sure there is safegaurds in place to prevent duplicates but i guess maybe goblin marriage will be weirder.

Additionally i noticed there was code that prevented surnames from being titles like " the wolves" or " of wolves". I placed a comment on where this code is if you ever want a marriage to share the same title.

## Why It's Good For The Game
If people want to ban gay marriage they could just do it in game. If 3 people are fine with the marriage then it should be allowed. Unless there is a solid canon lore that the divine creature officiating this union of souls hates gays.

If this also seems too easy of a way to get triumphs we can put in more hoops such as preventing nobles from marrying peasants or requiring a ring to be present since rings are made from a uncommon material and the idea of the king scouring the land for gold so he can marry seems humorous.

tweak: fluff.dm /obj/structure/fluff/psycross

## Pre-Merge Checklist

### DIFF
--- a/code/game/objects/structures/fluff.dm
+++ b/code/game/objects/structures/fluff.dm
@@ -1016,78 +1016,96 @@
 					return FALSE
 				var/marriage
 				var/obj/item/reagent_containers/food/snacks/grown/apple/A = W
+
+				//The MARRIAGE TEST BEGINS
 				if(A.bitten_names.len)
 					if(A.bitten_names.len == 2)
-						var/list/found_mobs = list()
+						//Groom provides the surname that the bride will take
+						var/mob/living/carbon/human/thegroom
+						var/mob/living/carbon/human/thebride
+						//Did anyone get cold feet on the wedding?
 						for(var/mob/M in viewers(src, 7))
 							testing("check [M]")
-							if(found_mobs.len >= 2)
+							if(thegroom && thebride)
 								break
 							if(!ishuman(M))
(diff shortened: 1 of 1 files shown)


BadAtThisGame302:
So... we've become... 
### **_Woke._**

sanshoom:
The #WOKEMOB is here. (this must be added)

InsightfulParasite:
Still testing to make sure things dont explode when people are married and it works like i described.

InsightfulParasite:
Im marrying Vivian of Rivia to Lilian to Givia to test the surname thing. Apparently if you have a title such as "of hellmires" then it will be thrown out.

InsightfulParasite:
I married a goblin to a werewolf i think that means it works.

zeratino:
@maaacha HEEEYYYYYY... LET'S GET MARRIED...

Stutternov (approved):

PR #221, opened 2024-06-20
Tweaks Padded Dress Crafting 

OctusGit:
## About The Pull Request

The padded dress being in sewing and being so cheap to make despite its insane stats and coverage feels like an oversight. It deserves way more work and material to make given it is the clothes of queens and royals. Its now in weaving, needs silk, and actually has a sellprice now.

## Why It's Good For The Game

Outfit of queens and royals being basic sewing is not good and reflective of its protection and quality.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/roguetown/roguecrafting/sewing.dm
+++ b/code/modules/roguetown/roguecrafting/sewing.dm
@@ -216,12 +216,3 @@
 	reqs = list(/obj/item/natural/fur = 2,
 				/obj/item/natural/hide = 4)
 	craftdiff = 5
-
-/datum/crafting_recipe/roguetown/weaving/armordress
-	name = "padded dress"
-	result = /obj/item/clothing/suit/roguetown/armor/armordress
-	reqs = list(/obj/item/natural/cloth = 6,
-				/obj/item/natural/fibers = 4,
-				/obj/item/natural/hide = 4)
-	craftdiff = 5
-	sellprice = 80
--- a/code/modules/roguetown/roguecrafting/weaving.dm
+++ b/code/modules/roguetown/roguecrafting/weaving.dm
@@ -117,3 +117,12 @@
 				/obj/item/natural/fur = 2)
 	craftdiff = 5
 	sellprice = 60
+
+/datum/crafting_recipe/roguetown/weaving/armordress
+	name = "padded dress"
+	result = /obj/item/clothing/suit/roguetown/armor/armordress
+	reqs = list(/obj/item/natural/silk = 4,
+				/obj/item/natural/fibers = 2,
+				/obj/item/natural/hide = 1)
+	craftdiff = 4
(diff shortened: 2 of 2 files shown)


ThePainkiller:
If it's the clothing of queens and kings, why is it on the blacksmith's closet?

code/modules/roguetown/roguecrafting/weaving.dm
```diff
@@ -117,3 +117,12 @@
 				/obj/item/natural/fur = 2)
 	craftdiff = 5
 	sellprice = 60
+
+/datum/crafting_recipe/roguetown/weaving/armordress
+	name = "padded dress"
+	result = /obj/item/clothing/suit/roguetown/armor/armordress
+	reqs = list(/obj/item/natural/cloth = 5,
+				/obj/item/natural/fibers = 2,
+				/obj/item/natural/silk = 5)
```
ThePainkiller, line 126: This is **not** royal clothing, and these are **padded** so it makes no sense to replace cloth with silk. At the very least cloth and hide make more sense than this. We **do** want more uses for silk, but this is definitely not one of them.

ThePainkiller (changes requested):
Silk? Not what things with armor values are made of.

code/modules/roguetown/roguecrafting/weaving.dm
OctusGit: Ok ive made changes, it now uses hide and has a higher material cost and crafting diff.

code/modules/roguetown/roguecrafting/sewing.dm
```diff
@@ -223,3 +216,12 @@
 	reqs = list(/obj/item/natural/fur = 2,
 				/obj/item/natural/hide = 4)
 	craftdiff = 5
+
+/datum/crafting_recipe/roguetown/weaving/armordress
+	name = "padded dress"
+	result = /obj/item/clothing/suit/roguetown/armor/armordress
+	reqs = list(/obj/item/natural/cloth = 6,
+				/obj/item/natural/fibers = 4,
+				/obj/item/natural/hide = 4)
```
ThePainkiller, line 225: 4 hides is insane. You do realize this dress was nerfed to the ground and you're making it literally not worth crafting over a leather/hide armor?
In fact, screw it, make it 4 silk 1 hide 2 fibers, since someone already changed the description to it being worn by nobility and the queen wears one, so at LEAST some silk, but for its armor values this is INSANE. Craftdiff is also insane, don't make it past 4 please. Nobody will want to do this shit if you make it so costly for so little return.

OctusGit: Ok thats fine I can do that

ThePainkiller (approved):

Stutternov (approved):

PR #243, opened 2024-06-24
Tanner Class

Max-023:
## About The Pull Request
My counterpart to the weaver, a towner role that works with leather rather than cloth. Good for making lighter armors as well as other clothing items. As such, the current items able to be made have been reassessed and made harder to craft for non-tanners and later down the line I may add more to the list.

The hunter has also been adjusted to more focus on being a proper hunter, as to not conflict with the tanner. Added cooking skill has been given, as well as polearms for flexibility in hunting.

## Why It's Good For The Game
Another crafter role that can offer more of an RP experience rather than fragging.

## Pre-Merge Checklist


Stutternov:
Going to let other devs review this. 

Imo - we have a lot of towner roles as is, the question is 'does this add enough to warrant de-specializing an airway underplayed role to add another?' I honestly think we should hold off on any more or less towner/drifter roles until every single one is evaluated with a plan of what is needed, wanted, and interesting.

NPC1314:
It seems the Tanner is better served being folded into the Butcher, adjust butcher skills if needed, no reason he shouldnt be the main leathercrafter as is.  Add random chance for him to spawn w papaka and whip maybe? Butcher doubling as leatherworker/tanner and maybe tiny bit medical is perfect to boost him a little IMO

Recipe changes for leather stuff looks good at a glance but Ill need confirmation from someone who crafts more ingame and got a better feel for cost balance of those items.

Hunter changes seems good if done combined with the above two steps, though roles do need a total oversight.

Optimism333:
I'm not sure if skincrafting as a system is robust enough currently to need a role dedicated to being specialized in it. The weaver role came with a loom and the addition of silk at least. I don't want to just shit on this idea, but I feel without some added complexity to tanning itself, like needing equipment and curing processes to do it rather than just a drying rack, there's not much point to this.

Dropping hunter's tanning from 3 to 1 seems rough. One must have item I usually craft as a hunter is a leather backpack, so I can carry more gear while hunting. If 1 is going to be too low to craft a backpack, that just seems like added tedium.

Stutternov:
Similarly - it's strange to increase craft difficulty on some things like the Papakha and Quiver; given the whip is lower effort to make for higher price. 

A papakha takes 1 fur (5 mammon value, drops rarer than hide noramlly) and 2 fiber (easy) for 10 mammon craft.

A whip takes 2 hide (10 mammon value) to make 39 mammon craft. Effectively more than triple the cost to make, yet lower difficultly adjustment.

ThePainkiller:
Hi, I'm the dev working on the "make towner less boring to play as" side of things currently

I was going to wait until map changes were merged and gucci so that I could develop an actual towner seamstress/tailor/clothier role, which will handle all three crafts equally expertly: Leathercrafter, Cloths, and Silk. They will be the light armor armorer slash drip clothier of the town, which will include their own workplace, as well as the mapping of loom and other required equipment for fast crafting of cloth and silk.

Leathercrafting alone does not warrant an entire job just for only one part of the whole crafting system, like Armorsmithing does.

PR #246, opened 2024-06-26
gives monks dodge expert, their own amulet; nerfs middle-aged stat-wise

saltprjak:
## About The Pull Request

Gives monks the dodge expert trait, and gives them their own (purely flavor) Ravox amulet. Middle-aged now gives one END at the cost of one SPD, as it was earlier in stonekeep.

## Why It's Good For The Game

Dodge Expert should give monks some much-needed combat sustain in line with other adventurers, without them having to rely on going Captain America with a shield. They don't have speed buffs, so they won't be consistently reaching 99% like other dodge classes do.  Psycrosses should in general be getting phased out of most loadouts, so they get their own amulet now, as well.
Middle-aged previously gave three stat points at no downside, a free advantage to anyone who knows of the oversight.
All changes approved by Macha.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/stats.dm
+++ b/code/modules/mob/living/stats.dm
@@ -51,9 +51,8 @@
 					change_stat(S, H.dna.species.specstats[S])
 		switch(H.age)
 			if(AGE_MIDDLEAGED)
-				change_stat("endurance", 1) //if a stat isn't listed then it dosen't have a bonus or negative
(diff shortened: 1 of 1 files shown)


Stutternov (approved):
other devs approved earlier; iffy on it given faith rework may rework this role a tad later but works for now at the very least.

PR #271, opened 2024-06-29
Thralls now have stamina

Max-023:
## About The Pull Request
This PR removes the trait that converted vampires get , the trait removing stamina as an issue.

## Why It's Good For The Game
Due to vampire thrall are expected to fight for their lord, it results in an unbalanced situation for the town as said thrall can infinitely swing and overwhelm their opponents without much skill or ease. Individuals with powerful skills like rogues could easily kill by swift stabbing at heads, ending fights quickly. This should make the regular thrall more on par with the towners they have to face, making fights a bit more enjoyable. Combat is delicately balanced with stamina in mind, so being able to infinitely swing is a huge problem.

## Pre-Merge Checklist


Plaguewalker:
Maybe instead of punishing thralls maybe make this pr for the VL instead? Thralls are meant to be the dime a dozen pawns, and in truth they can be put down rather quickly if knowing what to do and the guard isn't shitting or pissing themselves

Max-023:
> Maybe instead of punishing thralls maybe make this pr for the VL instead? Thralls are meant to be the dime a dozen pawns, and in truth they can be put down rather quickly if knowing what to do and the guard isn't shitting or pissing themselves. VL's need the nerf due to dagger spam along with sleep spells

If I were to do it to the VL, I would still apply it to the thralls. They still get a host of other benefits and right now the only way to kill them is to dogpile them else they will overwhelm you with the infinite stamina. This is reflected in the round where a rogue thrall murdered half the town within the first day as not even the guard could keep them in one location. If a 'dime a dozen' pawn can be so overwhelming, then they are over-tuned. 

That being said, I am also all for the VL getting stamina again. It makes no sense and is very messy that they get stamina, for the same reasons thrall do.

Stutternov:
If you can, prior suggestion to prevent sleep tiredness is a tad of a need for vampire.

If you can add that I 100% support this, it really is silly how sone antags/hostiles play. Good change.

Max-023:
> The no stam trait is also the trait that makes vampires not get the tired debuff, which I think is pretty essential as they are supposed to be creatures of the night. Would you be able to give them a lesser trait which simply stops them from getting the tired debuff? The code for getting tired is in the code/controllerssubsystem/nightshift file.

I could look into it, but I'll have to see if it is a possibility with my skills.

Max-023:
Aaaaaand it was a lot simpler than I imagined. Sweet.

Stutternov (approved):

PR #288, opened 2024-07-01
Elder Skill Additions and Adjustements

ToastyWiz:
image

## About The Pull Request

Gives the elder riding skill, as the new Dakka map gives him a saiga to ride round-start, much like the king. Very useful for both RP and him getting around as the top dog from the village to the town.

Also gives wrestling, because why didn't he have it as is? He already has no armour training or dodge expertise, letting him be grab bait is horrible. Did lower the fistfighting from 4 to 3, though, to make up for the 3 in wrestling. Puts him on par with other classes with combat skills.

Athletics is just a tiny bonus.
## Why It's Good For The Game
It makes the elder better feeling to play without really changing too much.

## Pre-Merge Checklist


AceMcLazer:
I dig it. Elder has always been meant to be like the village law with his cudgel and cells. So SOME wrestling makes sense since you don't want to frag when you can capture.

ToastyWiz:
Him not having wrestling makes actually taking on anything that you'd need to capture horrifying as you're one grab away from most pilgrim classes from being slammed into the ground.

Stutternov (approved):

PR #298, opened 2024-07-02
The Great Rebalance

DoctorRay2:
## About The Pull Request

Rebalances all job roles and drifter roles and changes some economic values and materials.

## Why It's Good For The Game

Currently most/all/ roles are combat focused, this PR changes the stat and skill values of most job roles and drifter roles, and reenables a few. It also changes the price value of some materials. It changes how smelting iron into steel works and now will only produce 1 steel bar instead of 3.


1N7Slayer:
unatomic PR [👎](https://emojipedia.org/thumbs-down)[👎](https://emojipedia.org/thumbs-down)[👎](https://emojipedia.org/thumbs-down)[👎](https://emojipedia.org/thumbs-down)[👎](https://emojipedia.org/thumbs-down)

Max-023:
Overall, I do think this is a good return to form. There are personal things I'd adjust, but those are only minor things. While this might not be entertained, it is definitely an interesting path we could take.

Stutternov:
Holy based.........

Optimism333:
While it is a draft, I have to say many of the changes in this pr currently are... bizarre. Very heavy handed, as one has said, and I think many need adjustments or further consideration.

- Amazons now have a greater strength boost than barbarians and get 3 in polearms, plus a 44 probability of an extra 2 skill, which lets them get up to 5 skill in polearms.
- Barbarians getting their strength, constitution, and endurance nerfed down to 1s across the board is a strange choice. I never considered them to be overly powerful, as far as the drifter classes went. They had middling combat stats and no armor training.
- Why reduce the acolyte's unarmed and wrestling skills down to 1? And add a strength malus of -1 onto that. 
- Why are the garrison guard eating shit in their combat stats? Such that they can roll as low as 1 in their weapon skills? That justification of "they're a poorly trained militia" doesn't match their gear or what their description describes them to be. They are trained and armored, they shouldn't have a chance of rolling less in their combat skills than a peasant carpenter.
- Manor guard changes are less egregious, though I do not believe they should have a chance to roll as low as 2 in their combat stats. They are knights now, they should be skilled in weapons they have training in at minimum.
- Soilsons probably did need some adjustment, but I do not think taking their polearm skill away completely was necessary. Same goes for peasant and legendary peasant.

PR #304, opened 2024-07-03
Nitemaiden Title Addition

Max-023:
## About The Pull Request
Adds a title for the male characters that is more befitting for them.

## Why It's Good For The Game
Because the big burly men offering shoulder rubs deserve representation.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/jobs.dm
+++ b/code/modules/jobs/jobs.dm
@@ -115,7 +115,7 @@ GLOBAL_LIST_INIT(peasant_positions, list(
 	"Grabber",
 	"Mercenary",
 	"Niteman",
-	"Nitebachelor
+	"Nitebachelor",
 	"Docker",
 	"Prisoner",
 	"Beggar",


code/modules/jobs/job_types/roguetown/peasants/bathmaid.dm
```diff
@@ -1,14 +1,16 @@
...
 	department_flag = PEASANTS
 	faction = "Station"
 	total_positions = 4
 	spawn_positions = 4
+	f_title = "Nitemaiden"
 
 	allowed_races = list(
 		"Humen",
 		"Elf",
+		"Dwarf",
```
BadAtThisGame302, line 13: This is unbelievably cursed.

Stutternov:
fails to compile, L

Max-023:
> fails to compile, L

Fixed now, nerd

code/modules/jobs/job_types/roguetown/peasants/bathmaid.dm
```diff
@@ -1,14 +1,16 @@
 /datum/job/roguetown/nitemaiden
-	title = "Nitemaiden"
+	title = "Nitebachelor"
```
ThePainkiller, line 2: Big yikes on using "bachelor" for a position about gucking.
During early and late medieval times, "bachelor" was reserved for young knights (which were not yet part of any order of chivalry) and young members of artisan or merchantry guilds. It not only feels weird to put this title on a whore, but this is also the equivalent of _"I just turned eighteen so I'm legal"_.
I understand _"maiden"_ is not an apt title for literal prostitutes as it implies virginity, so how about just naming them _"Bathmaids"_ instead? Hell even the job description paints them as _"bathwenches"_.

ThePainkiller (changes requested):

code/modules/jobs/job_types/roguetown/peasants/bathmaid.dm
Max-023: Maiden was also the term for young women as well, so I'm confused as to why now this is suddenly an issue. Regardless, I can look for terms better suiting the whores that is a bit more subtle.

ThePainkiller: It's not that it's "an issue", maiden literally meant back then "a virgin woman", which is counterintuitive considering they most definitely are not lmao
Like I said, bathmaid or bathwench is fine, think like Kingdom Come Deliverance.

code/modules/jobs/job_types/roguetown/peasants/bathmaid.dm
Wesreidau: Oy voold eou leek a sveeedish massaaage?

code/modules/jobs/job_types/roguetown/peasants/bathmaid.dm
Wesreidau: I agree 'maiden' doesn't make sense compared to something like nite-wench or bath-wench. The masculine would be nite-servant or bath-servant, imo.

DAKKA-WAAAGH: Could just call them whores or prostitutes, dunno why it needs to be complicated. It's a brothel, whores work at brothels.

Stutternov:
Hasn't been updated upon review; re-open if work continues outlined above.

PR #314, opened 2024-07-03
dakkatown graveyard and townhouses

YuiY1997:
restores some of the old graveyard and temple of Necra, adds nicer looking townhouses for skilled craftsmen, provides more public space for hanging out and chilling, adds keyes for each of the houses.

Should not change game balance, just adds flavor and gives stewards more property to give out.

I might experiment with making some towner start-spawns inside of the locked houses.

## About The Pull Request

Adds some of the features from the old graveyard

REpalces an empty, wooded lot with the waterfront district. New townhouses made for people richer than peasants but poorer than proper lords to be given out at the steward's discretion.

## Why It's Good For The Game

Gives more places for roleplay, adds flavor, adds detail and SOVL

## Pre-Merge Checklist

image

image

image


YuiY1997:
I am considering that in the future these houses could be given to important roles that would make sense to have their own property that they own personally.

For example captain, jester, veteran, or royal guards

Optimism333:
Town houses with KEYS? Bless you, sir

Stutternov (approved):

PR #340, opened 2024-07-07
Moves the Bathhouse back below the Inn

bowlofcereal:
## About The Pull Request

Takes the Bath House and puts it back under the Inn, rather than it be its own independent building. Removes the previous building and replaces it with some foliage until either someone else or I decide to fill it in with something-- which I intend to do  eventually.

## Why It's Good For The Game

The Bathhouse as it relatively stands is heavily isolated from the rest of the town, and for most of the round can and will be absolutely empty and unused besides someone breaking down the windows too loot it. In previous iterations it was located under the inn, but was moved to be its own separate building later. 

This PR adds a stairwell leading below the inn outside that leads into the bathhouse, bringing it closer to the town. No reason to not really have it like this-- when the player-base is at its highest pop it is only 36-44, which leads to the previously mentioned part of the map being desolate anyways. Makes more thematic sense for something that could be seen as controversial being underground and away from everyone as well.

Also this is in no way meant to replace the tavern-- I am hoping this actually _encourages_ the Barkeep and the Niteman to work a bit more closely and incentivize cooperation between roles.

Entrance
bathentrance

Bathhouse
bathhouse1

## Pre-Merge Checklist


bowlofcereal:
this is for Dakkatown btw if it wasnt obvious, might do it to the old map as well if we ever add rotation

NPC1314:
> this is for Dakkatown btw if it wasnt obvious, might do it to the old map as well if we ever add rotation

Yeah its why its marked as unapproved, more info will come in the mapping discord in the coming days.
Dakkatown is going back to dev phase so the old map will be the one played, I'd say make a PR for oldmap with a similar style since the basic idea seems good.

bowlofcereal:
on dakkas request, some ingame images and the full downstairs image

full image
image

rest of these below are from a localtest
bathhouse6
bathhouse5
bathhouse4
bathhouse3
bathhouse2


I tested all the keys and they all have the proper lockIDs as well. Everything does in fact work and nothing explodes, so we win these

Still open to change suggestions

Wesreidau:
I dislike moving the bathhouse to the center of town. Social hub or not, I think it belongs on the outskirts as the social hub of the ne'r-do-wells among the slum housing and next to its hidden smuggler's gate. Additionally this squeezes the whole building to a single Z-level, which I think is compressing too much.

DAKKA-WAAAGH:
I've had time to think about this PR and I have to say I'm not a fan of going back to the days of having the bathhouse under the inn. I understand your concerns, but you have to keep in mind this map was designed with a higher pop in mind from the get-go and most people trying the map were confused about many things -- not just where the brothel was. Changes like these are just too hasty.

After all, what exactly does moving everything closer to the center of town solve if it's meant to feel like a full town and not just one giant social hub blob? If the brothel is moved to the center of town then shouldn't the merchant join them as well since there were similar concerns? I don't think we should approach map design that way. In this case, flavor should come before mechanics, and the brothel fit well already in the seedy part of town where they could be well-insulated from the attention of the guard and the church to do _as they please_.

If you or anyone else wants to make changes to the brothel, then please make changes to the current one without completely removing the flavor from that part of town.

bowlofcereal:
going to close this on dakka's request to not move the bath house, ill just make edits to the area its currently in later

PR #366, opened 2024-07-12
Runtime slaying

NPC1314:
Good old 51 gets a bulet in the head. Works fine local, testmerge and see how it does on live server.

## Pre-Merge Checklist


Stutternov (approved):

NPC1314:
Was meant to be a TM only kinda thing, I probably should have labeled it to make it clearer, alas.

PR #375, opened 2024-07-14
Saint of Wine

NPC1314:
## About The Pull Request

Changes Eora focus from horny to marriage and wine making. Probably better in actual gameplay even if pretty hard to redeem now.
Let Guayo take a look before merge, deviates from opposing pairs in sins obviously (unless some gods goes teetotaller)

### DIFF
--- a/code/datums/gods/patrons/divine_pantheon.dm
+++ b/code/datums/gods/patrons/divine_pantheon.dm
@@ -92,13 +92,13 @@
 	worshippers = "Smiths, Miners, Artists."
 	sins = "Cheating, Shoddy work, Suicide"
 
-/datum/patron/divine/eora
+/datum/patron/divine/eora // should get some wine spell I guess
 	name = "Eora"
-	domain = "Saint of the Love, Life and Beauty"
-	desc = "Blind love for all, without a shred of hate in her heart"
-	flaws= "Naive, Compulsive, Ignorant"
-	worshippers = "Lovers, Harlots, Doting Grandparents"
-	sins = "Sadism, Celibacy, Ruining beauty"
+	domain = "Saint of Wine, Fertility and Beauty"
+	desc = "Wine and children warms the heart"
+	flaws= "Naive, Intoxicated, Vain"
+	worshippers = "Married couples, Drunkards, Vain people"
+	sins = "Wasting wine,  Ruining beauty, Violating marriage"
 	t0 = /obj/effect/proc_holder/spell/invoked/lesser_heal
 	t1 = /obj/effect/proc_holder/spell/invoked/bud
 	t2 = /obj/effect/proc_holder/spell/invoked/eoracurse


sanshoom:
this is such a random change that doesn't make any sense. people are still gonna treat her as the love goddess

BadAtThisGame302:
Why even do this change? Like people ERP and follow her and why change that? What is the the explanation behind removing that part for no reason

Stutternov:
Personally I think she's fine beyond her issue with chastity and worshipers being harlots. I think a focus on the love aspect and less on the screwing aspect would be beneficial.

However I do think something like tying her with wine and stuff as a spell would be interesting.

Mogelix:
I share in the sentiment that this change doesn't seem very thought through. It has a lot of implications and needs a strong implementation to not be to the overall detriment of the game's lore consistensy. At face value however, some of these are interesting to consider. Some things:

1. The change from dual opposite sins is clearly detrimental. It undercuts the conflict of the Pantheon lacking a single unifed morality. Maybe Noc could shun intoxication rather than carnal desire? This would conflict with existing headcanon I've noticed among Noc-worshipping characters players that Noc-worshippers love taking powders to 'uncover secret wisdom'. Also, it would erase the existing characterisation of the Gods with Noc as prude and Eora as whore.

2. This change would clear the way for the Church to act a lot more unreservedly Christian in its sexual morality (no sex before marriage) without an Eoran free-love ethos to oppose it. But this problem eludes the scope of lore, and is more a problem that reflects Church players and others defaulting to Christian morality in roleplay (rather than reflecting the virtues of their Patron like Astrata who is indifferent to sex that doesn't disrupt hierarchy/order.)

3. As things stand, I believe Dendor acolytes are the ones who can magically turn water into wine. Changing Eora to be more wine-focused would mean eventually giving this power to Eoran acolytes, and then also creating a new spell for Dendorites.

4. Eora's amulet is a heart with a tongue. It exudes horniness, sensuality, and sexuality rather than marital familiarity. This is just one example of the 'legacy' of horny Eora that would and should be considered/have to be reworked. 

5. Our google docs primer Gods information (which many still regularly use) remains out of date and does not reflect in-game 'official' Gods lore. Someone would need to update this eventually.
 
Overall, a lot of challenges and questions. One of them being: is having a 'horny' god really such a problem that needs to be solved in the first place? The Gods are one of the few 'concrete' things in the lore, so changing them in my opinion needs this kind of consideration and a strong implementation to prevent players completely falling back on an outdated collective headcanon.

CormanQuewye:
I'm not certain what the point of reworking Eoran horniness is besides 'not in my christian SS13 server'. People who ERP will continue to do so even if Eora is changed (not all ERPers are even Eoran). The Gods are explicity left open, so the Eoran priesthood would, again, still act in such a way. It's so weird to me that in an openly adult server where people try to be nasty and brutal and stab each other for fun that 'goddess says pintil in hole good' is where they draw the line.

If the clergy needs any rework it's giving them a purpose to exist beyond Feldsher plus. I've mentioned it before, but give the Eoran priesthood a cleanliness spell so they can be the town janitors.

NPC1314:
Hmmm interesting feedback, for context I'm the one who came up with the sin concept and sat down with Guayo doing the pairs etc so I'm very familiar with the setup, breaking it is probably a bad idea though.

I don't give a rats ass about the moral aspect, goon all day for all I care, but there are such things as incentives and repulsion factors, the people do X anyways doesn't carry water since when/where/how matters and change with context.  Church roles should have clear incentives to not spend all round hiding and gooning and with how Eora turned out its a tall order to motivate chastising a priests for that sort of thing while still respecting the lore.

Beyond having opposing sin pairs the idea was also to have something practical to tie in, but Eora ended up having just the gooning, which is a bit boring. Wine bit obviously a Dionysus/Bacchus nod.

The whole point of structuring up and pushing the lore ingame was to reduce headcanon clashes while still keeping this somewhat vague and meshing with gameplay, but either way, its not something that can be sorted in a small PR like this.
I'll let Guayo ponder this while I'm sipping vodka on the beach, thanks for playing folks.

PR #402, opened 2024-07-19
Aasimar stat change

Marmio64:
## About The Pull Request

Changes aasimar stats to the following-

Base stats

+1 strength
-1 perception
-1 intelligence
-1 constitution
+1 endurance
0 speed
-1 fortune

Men get the standard +1 constitution and endurance and women get +1 intelligence and speed.

## Why It's Good For The Game

Giving aasimar a unique niche of being sort of a glass cannon, but reckless race. 
They have STRENGTH given by the gods and they have the ENDURANCE to be reckless, but due to being reckless they lack PERCEPTION and INTELLIGENCE, while also their abandonment left them without FORTUNE and long years sapped their CONSTITUTION.

I used the model of races having a net stat bonus of +2 to make this, though also going under the idea that strength is worth 2 points and perception and fortune are both only worth 0.5 points since they don't do much of anything. As a result, under my theory it looks like-
+1 strength (2 points)
-1 perception (1.5 points)
-1 intelligence (0.5 points)
-1 constitution (-0.5 points)
+1 endurance ( 0.5 points)
0 speed (0.5 points)
-1 fortune (0 points)

+2 points from male/female bonuses to a total +2 point

Yeah this is just random balancing done by someone who doesn't play the race and isn't a maintainer either but really just my two cents on what I think may be interesting. Close or keep it, however you want.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/roguetown/other/aasimar.dm
+++ b/code/modules/mob/living/carbon/human/species_types/roguetown/other/aasimar.dm
@@ -44,8 +44,8 @@
 	OFFSET_FACE_F = list(0,-1), OFFSET_BELT_F = list(0,-1), OFFSET_BACK_F = list(0,-1), \
 	OFFSET_NECK_F = list(0,-1), OFFSET_MOUTH_F = list(0,-1), OFFSET_PANTS_F = list(0,0), \
 	OFFSET_SHIRT_F = list(0,0), OFFSET_ARMOR_F = list(0,0), OFFSET_UNDIES_F = list(0,0))
-	specstats = list("strength" = 0, "perception" = 1, "intelligence" = 1, "constitution" = 0, "endurance" = 1, "speed" = 0, "fortune" = 1)
-	specstats_f = list("strength" = 0, "perception" = 1, "intelligence" = 2, "constitution" = -1, "endurance" = 1, "speed" = 1, "fortune" = 1)
+	specstats = list("strength" = 1, "perception" = -1, "intelligence" = -1, "constitution" = 0, "endurance" = 2, "speed" = 0, "fortune" = -1)
+	specstats_f = list("strength" = 1, "perception" = -1, "intelligence" = 0, "constitution" = -1, "endurance" = 1, "speed" = 1, "fortune" = -1)
 	enflamed_icon = "widefire"
 	patreon_req = 0
 


Optimism333:
I don't like the hit to constitution, think it should be the opposite given that they were molded from the soil to be soldiers.

The rest looks fine, though personally I would up the INT malus to -2, otherwise picking female for the boost of +1 intelligence will negate it entirely.

sanshoom:
The hit to constitution really doesn’t make sense tbh. They were made for war, so why wouldn’t they be durable?

the term ‘glass cannon’ doesn’t make sense for this either lol

Optimism333:
Well, alright then. If that's what you want, and you have no intention of making changes outside of a maintainer asking for them... sure.

I will say I disagree with these stats in their current form, and I don't think it matches what their lore and visual references indicate them to be.

ThePainkiller:
Be evil. Give END and CON but take away -1 SPD, with zero touching their INT >:)

nvm, Fortune AND INT malus actually makes them impossible to craft shit that's even more evil than Speed debuff (but the fortune debuff makes sense lore wise...)

Optimism333:
I'm actually down with aasimar having a fortune debuff, and the reason I suggested INT being debuffed to -2 in the suggestion I opened about this is a month ago because that's a pretty meaningful debuff to the ability to craft outside of trained skills.

-1 INT like the current pr proposes isn't great, but picking female for +1 just negates that debuff. It has to be at least -2 to be felt across both male and female aasimar.

assortedbeads:
We have orcs at home
Orcs at home:

Stutternov:
Other stat change was done; for now going to see how that one goes as it's a tad less radical compared to this stat layout.

Optimism333:
> They're soldiers created by god, imagine having penalties against knockdown and not being able to break from grabs more easily

I personally prefer them having strength and constitution as their boons, rather than strength and endurance.

Being the longest lasting stamina-wise in a fight can remain the domain of dwarves, but I think aasimar being sturdier than average fits them well as soldiers.

More aloof about the idea of swapping their negatives around, though I think the heavy hit to intelligence I suggested in my pr makes sense given their lore.

ThePainkiller:
I want to eliminate tackle meta so badly tbh

ThePainkiller:
Good to close this PR if no more changes will be introduced?

PR #436, opened 2024-07-27
return from break: return of the recurve bow

cornercube:
## About The Pull Request
adds the recurve bow
deals slightly less damage, but charges shots faster and needs less stamina to keep held

adds damfactor to bows and crossbows
multiplier to ranged damage
recurve bows have 0.9 damfactor, so damage from ammo is reduced by 10%
longbow has a damfactor of 1.2, so 20% more damage (25 force would only make it a more effective club)

the shop can import recurve bows for 60 coins

## Why It's Good For The Game
ranged variety

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/apprentices/prince.dm
+++ b/code/modules/jobs/job_types/roguetown/apprentices/prince.dm
@@ -22,6 +22,7 @@
 
 /datum/outfit/job/roguetown/prince/pre_equip(mob/living/carbon/human/H)
 	..()
+	H.virginity = TRUE
 	if(H.gender == MALE)
 		pants = /obj/item/clothing/under/roguetown/tights
 		shirt = /obj/item/clothing/suit/roguetown/shirt/undershirt/guard
@@ -32,7 +33,6 @@
 		beltr = /obj/item/storage/belt/rogue/pouch/coins/rich
 		backr = /obj/item/storage/backpack/rogue/satchel
 		if(H.mind)
-			H.virginity = TRUE
 			H.mind.adjust_skillrank(/datum/skill/combat/axesmaces, 1, TRUE)
 			H.mind.adjust_skillrank(/datum/skill/combat/bows, 3, TRUE)
 			H.mind.adjust_skillrank(/datum/skill/combat/crossbows, 1, TRUE)


ThePainkiller:
Localtest and post screenshots of bow damage variance, please. 

And my suggestion: damage multipliers on ranged weapons would be better defined at "launcher" parent, so that crossbows may benefit from it (for future wrist-mounted crossbow idea and such)

Also I believe more damage penalty will be more impactful than 10% less damage, which will barely be felt (a 20 damage arrow would do 18)

cornercube:
> Localtest and post screenshots of bow damage variance, please.
> 
> And my suggestion: damage multipliers on ranged weapons would be better defined at "launcher" parent, so that crossbows may benefit from it (for future wrist-mounted crossbow idea and such)
> 
> Also I believe more damage penalty will be more impactful than 10% less damage, which will barely be felt (a 20 damage arrow would do 18)

hilderp

with PER damage modifier, the difference in damage is about 3-4
above was tested with 11 PER, 30 vs 33
though the recurve bow is meant to be overall similar to the regular bow, with the reduced charge time not likely to matter much

crossbows are not children of bows, and both of them have their damage defined individually
so i just added the same line of code to crossbows

ThePainkiller:
> with PER damage modifier, the difference in damage is about 3-4 above was tested with 11 PER, 30 vs 33 though the recurve bow is meant to be overall similar to the regular bow, with the reduced charge time not likely to matter much

Yeah those numbers aren't muuuch of a difference, that was exactly my point

> crossbows are not children of bows, and both of them have their damage defined individually since crossbows do not have damage modified by PER and probably shouldnt be defined by the parent so i just added the same line of code to crossbows and gave them damfactor as well

You apparently have missunderstood my comment. Bows and Crossbows are children of grenade launcher revolver type. Their path is both `revolver/grenadelauncher`. What I meant to say is to put the variable and checks on `grenadelauncher` parent so that it is shared for both.

cornercube:
> Yeah those numbers aren't muuuch of a difference

in any case, its not meant to be a huge departure from the regular bow 

> What I meant to say is to put the variable and checks on grenadelauncher parent so that it is shared for both.

they have different damage modifiers, the crossbow doesnt get a boost from PER
so you probably dont want it on the grenade launcher path anyways

Stutternov:
I can say this code has worked elsewhere and seems OK, I would enjoy if maybe the bow can be made a bit more unique but I'm actually happy with how it is currently; archer flavor based flavor.

ThePainkiller (approved):
Good to merge then, we'll see how they play out through player feedback and adjust if necessary

PR #441, opened 2024-07-27
RogueHamlet minor nitpick tweaks

bowlofcereal:
## About The Pull Request

Edit a few things that I noticed would be an issue or things I did not like

Spawn landmarks - there were no landmarks for the prince, it still had orphan landmarks/servant landmarks (those are removed now), added a puritan landmark
A few random ugly stains on the map (tables with two different sprites used)

## Why It's Good For The Game

Making the map not have a ton of issues with spawns when we playtest it, otherwise I will be doing some tweaks here and there to specific areas of the map later on-- but this is a generally good start

## Pre-Merge Checklist


bowlofcereal:
never mind #443 exists and is better.. bleak

PR #442, opened 2024-07-27
Assassin Changes/Nerf

DovaKitty:
## About The Pull Request
* Lowers Drifter Assassin speed stat by 1 and strength by 1. This brings them closer in-line with Rogue.
* The profane dagger now does 14 damage vs non-Hunted targets, and 20 to Hunted targets.
* Assassin is now an antag, which can be tracked and assigned. Drifter assassin starts as an assassin antag.
* Non-drifter assassins that are assigned the antag role midround are revealed to be Graggar cultists in disguise. They can retrieve their profane dagger from statues among other places.
* Assassins get the NOSTINK and VILLAIN traits.
* Assassins get new combat music.
* Drifter Humen Assassins get unique laugh and whistle noises. Use them to strike terror into your victims before their final end.
* hunted.log has been added, which stores all the Hunted of the round.
* Admins can now use the 'Mark for Assassins' punishment, which marks the character for death and notifies all current assassins.

## Why It's Good For The Game
Brings Assassin more in line with the standard power-curve of the server. They will still be able to kill their targets quickly with 20 damage swift strikes, but will be comparable with Rogue when fighting other opponents.

## Pre-Merge Checklist


sanshoom:
Scream if you love Graggar!

code/modules/antagonists/roguetown/villain/gassassin.dm
BadAtThisGame302, line 1: Should probably rename the maniac villian.dm to maniac instead of there having two similar assassin.dms

sanshoom:
Is there any reason why those laughs and whistles have to be only for humens?

code/game/objects/items/rogueweapons/melee/knives.dm
```diff
@@ -254,7 +254,7 @@
 /obj/item/rogueweapon/huntingknife/idagger/steel/profane
 	name = "profane dagger"
 	desc = "A dagger made of cursed black steel. Whispers emanate from the gem on its hilt."
-	force = 20 // Very powerful for a dagger, but still below a two-handed sword by a large margin. Fitting for an enchanted weapon.
+	force = 14 // Now does 14 damage to non-hunted victims, but 20 to hunted victims. Sic Semper Gragganis.
```
ThePainkiller, line 257: No need to define the var as 14 since that's steel dagger damage and it will be inherited :P

ThePainkiller:
May I suggest adding the trait Steelhearted to assassin? They shouldn't feel bummed by watching people get delimbed

code/game/objects/items/rogueweapons/melee/knives.dm
DovaKitty: Agreed, if the steel dagger is ever changed damaged wise, this would stop that. Changed.

DovaKitty:
> May I suggest adding the trait Steelhearted to assassin? They shouldn't feel bummed by watching people get delimbed

I thought they already had it. Slipped my mind. Added.

DovaKitty:
> Is there any reason why those laughs and whistles have to be only for humens?

It's standard as elves/dwarves have different voicepacks as Humens. So Assassin is on the same level as Elv/Dwarf, with Humen being the default. You can look at court magician and it's the same situation.

code/modules/antagonists/roguetown/villain/gassassin.dm
DovaKitty: Done, tested and it compiles and lets me in the game so it was easier to do than I thought.

ThePainkiller (approved):
Women hater perk might be a tad overpowered considering he gets "extra damage" on top of the dagger but eh, we'll see

PR #461, opened 2024-08-01
Movement Speed Granulation

DovaKitty:
## About The Pull Request
* Ports https://github.com/Rotwood-Vale/Ratwood-Keep/pull/502 and also adds armor weight affecting movement speed slightly.
* Movement speed is now affected per point of speed, rather than using 6 SPD and 14 SPD thresholds.
* Wearing medium or heavy armor will now slow you down by the equivalent of -1 or -2 SPD respectively.
* Wearing armor you are not trained to wear will double the above penalty.
* Modification to speed remains small, only being noticeable over larger runs. For reference, the maximum penalty this code imposes is '1', where going into sneak mode is '6'.

## Why It's Good For The Game
Currently, SPD does not affect movement speed unless it brings you over the threshold of 6 or 14, and then does so very jarringly. This means most characters will not have their SPD stat affect their movement whatsoever. This PR fixes that and adds a very small penalty to wearing heavier armor. Currently, there are few downsides to wearing heavy armor if you have the training required. This PR will make unarmored characters slightly faster on average than heavily armored characters.

## Pre-Merge Checklist


ThePainkiller:
Tested how fast this makes high-speed classes?

DovaKitty:
> Tested how fast this makes high-speed classes?

Tested it ingame. Bonuses cap out at 15 SPD and reflect a 20% bonus when compared to 10 SPD. This is assuming they are not wearing any medium/heavier armor. If this needs to be reduced, it can be, but that's the maximum and can't really be reached by 95% of class/racial combos.

ThePainkiller (approved):
Makes speed actually matter by the point and not at very specific thresholds, therefore good

PR #466, opened 2024-08-03
Expands character slots to 20.

1N7Slayer:
## About The Pull Request

Expands character slots to 20 from 10. Chat, is this real?
1 line code change lol. Have fun with it.

## Why It's Good For The Game

I often felt very limited to 10 character concepts. I don't wanna overwrite or delete characters for roles I may play in the future but might also not. To solve this conundrum of my own which I believe to also not be exclusive to me, I made this PR.

## Pre-Merge Checklist


BadAtThisGame302:
Maybe 20 is better than 30. More limiting but not as much as 10. Still shouldn't have too much.

Stutternov:
Prob put it to 20, this can cause strain after long periods of time on a database as every unique Ckey causes entry.

Stutternov (approved):

PR #467, opened 2024-08-04
Convenience change to let admins send people back to the lobby easier, especially for already living characters.

1N7Slayer:
## About The Pull Request

Ported from https://github.com/Rotwood-Vale/Ratwood-Keep/pull/50.

## Why It's Good For The Game
Admin tools being usable and convenient is good, actually.

## Pre-Merge Checklist


ThePainkiller (approved):
If there's one thing we need, is more tools for our admins to have an easier job.

PR #469, opened 2024-08-04
Revenge of the Forgotten

DovaKitty:
## About The Pull Request
* Replaces the Puritan with the Inquisitor, a guest from Grenzelhoft that has been sent to help Rockhill with its monsters and cultists. What he lacks in magic he makes up for with knowledge and skill.
* Adepts do not serve the Forgotten God as readily as the Inquisitor as recent hires. They have lost their ability to cast spells and have gained in reward a needle, a silver dagger and one intelligence.
* The Forgotten God's worship now stretches to the Isle of Enigma. Those hailing from Grenzelhoft, including mercenaries, Counts, Inquisitors and Swordmasters worship the Last Hero.
* Grenzel titles of knighthood are added. Instead of 'Ser', those in the East use the title 'Retter', being a cross for the words 'Rider' and 'Savior' in the old tongue. Swordmaster and Inquisitor start with this confusing title.
* Torture has never been so fun! Torture mechanics have been fixed, allowing you to wrench sin out of both heretics and villains. Make sure to rough up your targets before you take them onto the pain train.
* Confessions have been reworked. Offering a confession to a heretic or villain will allow them to sign it voluntarily. Holding a confession in your other hand while you torture a heretic or villain will let the magical ink do its work and sign the confession on their behalf. Put these signed confessions in a mailbox as an Inquisitor to complete your objective and gain you and your adepts a TRIUMPH!
* Silver weapons have been added and standardized. While doing less damage and having less durability than steel, silver weapons are the BANE of vampires. In the future they will be more useful on more enemies. Ask your local weaponsmith to craft you some, and make sure to bring them silver ore.
* Grenzelhofter mercenary has had their dodge expert trait removed and their armor now counts rightly as medium armor. There was not much reason to give a class with +0 SPD super-special light armor and a trait for unarmored fighters.

## Why It's Good For The Game
Introduces a proper antag-hunting class, with some unique backing and mechanics. Makes the world a little more interesting and flavorful.

## Pre-Merge Checklist

- [x] You tested this on a local server.
- [x] This code did not runtime during testing.
- [x] You documented all of your changes.

PR #486, opened 2024-08-07
Old Psydonian Speech for Grenzelhoftians

Marmio64:
## About The Pull Request

Adds a unique old psydonain language available to only Grenzelhoft roles. Mostly designed to be put into use alongside #469 

Grenzelhoft roles would be all the ones defined in that PR-

Inquisitor/Puritan
Adepts/Confessors
Swordmasters
Grenzelhoft Mercs
Grenzelhoft Counts

Old Psydonian is an ancient language from before the Grenzelhoft empire had ever split, and is known chiefly by the clergy and inquisitional lodges.  Noblemen and devout soldiers often take lessons in the language as part of their theological classes, hoping to someday bring back the greatness of their former empire.

example of the font and language gibberish-
image

## Why It's Good For The Game

I think it would be interesting to make Grenzelhoftian roles seem more cohesive and solid, like they're all part of the same faction and speak the same language. In addition, allowing them to scheme together seems interesting and might open up unique opportunities, perhaps a dedicated grenzelhoft spy antag in the future?

## Pre-Merge Checklist

### DIFF
--- a/code/__DEFINES/lighting.dm
+++ b/code/__DEFINES/lighting.dm
@@ -85,7 +85,8 @@
 #define LIGHT_RANGE_FIRE		3
 
 #define LIGHTING_PLANE_ALPHA_VISIBLE 255
-#define LIGHTING_PLANE_ALPHA_NV_TRAIT 245
+#define LIGHTING_PLANE_ALPHA_LESSER_NV_TRAIT 236
+#define LIGHTING_PLANE_ALPHA_NV_TRAIT 222
 #define LIGHTING_PLANE_ALPHA_MOSTLY_VISIBLE 192
 /// For lighting alpha, small amounts lead to big changes. even at 128 its hard to figure out what is dark and what is light, at 64 you almost can't even tell.
 #define LIGHTING_PLANE_ALPHA_MOSTLY_INVISIBLE 128
--- a/code/datums/character_flaw/addiction.dm
+++ b/code/datums/character_flaw/addiction.dm
@@ -87,7 +87,7 @@
 	name = "Servant of Fire"
 	desc = "The warmth and just seeing something turn to ash is so much fun!"
 	time = 10 MINUTES
-	needsate_text = "I need to see something turn to ash, or be at fire. Anything!"
+	needsate_text = "I need to see something turn to ash, or be on fire. Anything!"
 
 /// KLEPTOMANIAC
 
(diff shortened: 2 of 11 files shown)


ThePainkiller:
This can either go well for characterization, or horribly wrong/cringy for very, **very** obvious reasons.

I like it, and I'd no trouble approving, but I'd like other devs to chime in and judge too because, well, you know how it usually goes when including german as a language

guayitoo:
Hey, Guayo here!

I don't think Grenzelhoft requires it's own language.

Alternatively, would you think that perhaps a stand in for a latin, perhaps even calling it Old Psydonic would be a good stand-in? Perhaps known only by the church and it's affiliates?

Likewise, a Zybantu language would be more appropriate considering their cultural differences is far more significant!

Let me know what you think!

guayitoo:
> > Hey, Guayo here!
> > I don't think Grenzelhoft requires it's own language.
> > Alternatively, would you think that perhaps a stand in for a latin, perhaps even calling it Old Psydonic would be a good stand-in? Perhaps known only by the church and it's affiliates?
> > Likewise, a Zybantu language would be more appropriate considering their cultural differences is far more significant!
> > Let me know what you think!
> 
> I think it's neat but also having it known by the Church seems kind of weird since they don't like to talk about psydon, and a zybantu language would be even more limited since only a single zybantine role exists.
> 
> I think Grenzelhoft really should have a unique language and be more and more integrated as a somewhat hostile but at the same time helpful faction since we already are getting some cool dynamics with visiting counts and the inquisitorial lodge. At the moment there isn't really any comparable outside forces other than grenzelhoft because the other factions only really have a single role associated with them (zybantine merc, elven sentinels/bladesingers, Dwarven longbeards) so to deny them more unique things seems like a lost opportunity.

We plan on expanding the Zybantu influence with more unique subcategories similar to Zybantu themed mercs, we're hoping to do this for things like Inquisitor, Merchant and other roles! So I don't think it's a waste.

Remember that, at least in lore, Grenz and Zybantu BOTH have big influence on the isle of Enigma. It may not reflect that in-game currently, but it is the plan to do so moving forward, you'd be contributing by committing to the idea as I've listed above.

As for why Grenz would use something like "Old Psydonic"? They used to be a mega empire similar to the Roman Empire back in the day, before collapsing and fracturing, the Grenzelhoft Empire is more reminiscent of the HRE trying to recuperate a lost glory.

Meanwhile Zybantu has always been more unified due to tyrannical methods and the fact that culturally almost all the rulers rejected the Pantheon, therefore, they should have a unique language to associate that foreignship from the mainland.

Hope this helps!

ThePainkiller (approved):
Approved by Loreholder, and that font goes hard af.

PR #475, opened 2024-08-06
Tweak devotion and clerics system

Animusin:
## About The Pull Request

This PR changes the devotion system. 
Clerical spells are tied to the level of holy skill. The experience is gained for spell casts.
The maximum devotion is 100 and increases by 50 for each level of clerical power.
All clerical spells now consume devotion as intended.
Devotion is gained from prayers. Having the Pantheon cross or neck cross lying nearby increases the amount gained.

Noc cleric gained invisibility and time-stop spells.
Eora cleric gained the mind swap spell.
Templars gained spells (discussable).

Major priests can be not only of Astrata. They have all spells of their domain from the start.

Removed Pestra from available patrons for clerics.

## Why It's Good For The Game

It's good for the game since it discourages being afk for getting all clerical spells. It adds some amount of interactability for clerics, as they need to have a cross nearby to gain devotion. And they can't cast their spells infinitely in a short amount of time.
Having major clerics of all domains makes the game more diverse.

## Pre-Merge Checklist

- [x] You tested this on a local server.
- [x] This code did not runtime during testing.
- [x] You documented all of your changes.

PR #504, opened 2024-08-11
Enable Stealth Mechanics

saltprjak:
## About The Pull Request

Comments out code that prevents sneak invisibility from updating. You will now go invisible in darkness while on sneak, as with any other server. Tested, both sneaking and looking around to find sneakers works. This also unfucks Noc's invisibility spell, which should let #475 merge without problems. Refresher: it was broken because the invisibility never wore off.

## Why It's Good For The Game

puts the ROGUE back in ROGUETOWN. iirc someone on the team wanted to improve(?) the stealth system way back in the beginning of Stonekeep and disabled it, but never got around to actually doing that, or re-enabling it. Right now the only issue I could find via testing is that the fade-in (exiting stealth) sometimes doesn't "play", resulting in you, say, appearing out of blank space a second later after leaving darkness, rather than gradually becoming less transparent. There's no way to remain permanently invisible tho, and you still fade out into stealth as normal.
image
image

## Pre-Merge Checklist


AceVX:
CRIME IS BACK IN A BIG WAY 🎉🎉🎉🎉

ThePainkiller:
I'd say we testmerge this one until people start noticing some issues with fade in / fade out speed.

Also we should totally make player's Perception handle how good they can stealth, a skill level on sneak vs player PER would be fine. I can see people abusing this against town guards with visored sallets or anything impairing vision to fuck them up without them ever being able to see them since they would be invisible on the reduced FoV despite being on full vision.

Another big important TODO is making this skill leveable as long as you aren't spotted while on stealth mode, per tick.

NPC1314:
Feels like its time to merge this today

Stutternov (approved):

PR #510, opened 2024-08-12
Garrison Skill Changes

beepbopbeepbop:
## About The Pull Request
Gives the Captain and Dungeoneer a proper skills buff they definitely deserve, changes up the skills of other garrison roles, including giving the Archer and Veteran some survivalist skills to demonstrate their role as a scout and a veteran of conflicts and campaigns respectively. Also, removes the cudgel the Captain spawns with as they are easily found in the armoury in abundance, and gives them the same boots as the Royal Guards, befitting of their status.

## Why It's Good For The Game
While garrison jobs may be changed up in the near future, I believe these changes will serve as a good foundation for such progress, and will address the deprecated skills of the Captain and Dungeoneer which have definitely suffered.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/dungeoneer.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/dungeoneer.dm
@@ -41,11 +41,11 @@
 	backr = /obj/item/storage/backpack/rogue/satchel	// lack of satchel requires dealing with the merchant to correct, which requires entering town; not ideal. N.
 
 	if(H.mind)
-		H.mind.adjust_skillrank(/datum/skill/combat/whipsflails, 5, TRUE) // for torture
-		H.mind.adjust_skillrank(/datum/skill/combat/axesmaces, 4, TRUE) // for executions w/ an axe
+		H.mind.adjust_skillrank(/datum/skill/combat/whipsflails, 4, TRUE) // for torture
+		H.mind.adjust_skillrank(/datum/skill/combat/axesmaces, 3, TRUE) // for executions w/ an axe
 		H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 4, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/combat/unarmed, 4, TRUE)
-		H.mind.adjust_skillrank(/datum/skill/combat/swords, 4, TRUE) // for executions w/ the big sword
+		H.mind.adjust_skillrank(/datum/skill/combat/swords, 2, TRUE) // for executions w/ the big sword
 		H.mind.adjust_skillrank(/datum/skill/misc/athletics, 2, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/swimming, 1, TRUE)
 		H.mind.adjust_skillrank(/datum/skill/misc/reading, 1, TRUE) // Allow reading notes passed to the literate noble prisoner, or writing reports. N. See peasants\prisoner.dm.
--- a/code/modules/jobs/job_types/roguetown/garrison/garrisonarcher.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/garrisonarcher.dm
@@ -50,10 +50,11 @@
 		head = /obj/item/clothing/head/roguetown/roguehood/red
 
 	if(H.mind)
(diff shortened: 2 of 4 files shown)


Stutternov:
Why nerf guards mace skill when their cudgel, the weapon intended for subduing people without murdering them, uses said skill?

Same kinda for the archer; I'd argue 1 sword skill is fully worth only having level 1 crafting and tanning just to not have to grind. And their wrestling is the only LTL skill you have; this just means to murder people more. Monks already have very high wrestling.

Optimism333:
I don't understand the method behind most of these changes. I thought we had a fairly decent garrison tier list as far as weapon training went before:

Garrison Guard (Skilled) -> Royal Guard (Expert) -> Captain/Veteran (Master)

One could argue where dungeoneer fell exactly, I would say in the garrison guard tier, due to the description saying that the dungeoneer is a former garrison guard who was taken on as warden, torturer, and executioner.

But yeah, most of these changes just seem to muddle distinctions and powercreep dungeoneer and royal guards into the master tier. Not a fan of the pr, such as it is.

beepbopbeepbop:
The things brought up by Stutternov are completely reasonable, although I do have some explanations for my thinking on certain things:

1. The dungeoneer receiving higher skill levels represents the fact they are experienced in using these tools of execution and torture, where previously they were absolutely worse than a garrison guard in just about every aspect despite the idea you say they were within the same tier, and could have even been a garrison guard before their position change, and let us keep in mind they have no armour training. I don't have an issue with lowering the big three of their combat traits, but I think giving them expert in whips & flails is reasonable because they are the only person in the game area using whips on a regular basis.
2. Royal Guards receiving mastery in polearms reflects the Medieval idea that polearms are the standard type of weapon, ranging from simpler designs used by levies, all the way up to the more advanced designs of the bec de corbin, while swords have mostly been used as secondary, defensive weapons. As the Royal Guards are knights, along with the Captain, I think it is proper that they would receive a great deal of training in the type of weapon they would be holding most of the time, far above anyone else. That being said, the issues with Royal Guard were really with their other stats, and I am not too concerned about them having expert polearms as that is good regardless, so I can agree to that.
3. Overall, most of the changes in the PR affected removing/adding skills that certain roles didn't have, and with redoing their non-combative skills; the changes to their main combat skills were only a small part that doesn't matter to me, so I have been happy to change it. The idea that the entire PR is not worth adding despite the fact these are changes that I could easily remove is a bit annoying, but they are fixed now.

ThePainkiller:
I don't know how to tell you this but the dungeoneer is not a combat class, will not ever be a combat class, and will never likely fit anywhere in skill levels halfway between a town guard and a royal guard while these other two roles gets nerfs for absolute zero reason.
The Mayor is not, and will not either, be a combat role. He's a diplomat, not a pseudo royal guard for the Soilsons. Stop trying to make him a combat class just because he has some old ass relic sword and a pseudo armory. Their combat skills should be warrior pilgrim-tier and never beyond that (I'd prefer Sword 2 then Knives 1 than the other way around.)
Royal Guard is fine as-is, as it's meant to be the paragon of skill for a two slot limited job which has the express job of making sure the king doesn't fucking die, so they have the equipment, skills and salary to reflect that. That's why they get both crossbow and bows, so they can choose either to fulfill an archer/knight role rather than both of them being purely melee classes. But the Veteran nerfs have no fucking ground. Yes, their flavor text states that they really want to retire already, but that's no justification to switch combat skills for living skills.
I don't even know where to start with Captain. They're the Head of Security adjacent job and you're nerfing their starting armor, tweaking combat skills in ways that make no sense other than just load them with options, and nerfing the fuck out of their athletics skill which will make them even worse at stamina cardio than a town guard (there's a reason they have it so high)

This PR feels like it has no solid bases to establish a hierarchy and eskews balance in ways I can't even make sense out of from a hierarchy perspective.

NPC1314:
This PR will be obsolete with the two PRs establishing a baseline and class selection for guards, it also has some core issues discussed in the comments above, so its time to close it.

My own reflection is a clear table of detailed changes in PR description is more important on skill change PRs than most since each point can make a pretty big difference and it should mesh with the overall design. My suggestion is to use a format like in #538  for similar PRs in the future.

PR #518, opened 2024-08-14
Riding skill effects speed (more so than it does already)

Marmio64:
## About The Pull Request

In addition to the standard large speed boost provided just by having any amount of riding skill, you also benefit an additional amount of speed depending on how much riding skill you have. The amount of bonus speed caps at expert level, though master/legendary riders still retain their other buffs for riding.

Tested it to make sure the speed buffs are noticeable but also will not make you stupid fast.

Maybe test merge this before merging? I feel like the speed buff is reasonable, but who knows what other people think.

## Why It's Good For The Game

Makes riding skill more useful. Also, Saigas could use the extra love.

## Pre-Merge Checklist


AceVX:
I would die of joy if we got functional mounted combat, lance charges etc. there's a jousting.dm but that doesn't appear to be roguetown to my knowledge

ThePainkiller (approved):

PR #525, opened 2024-08-15
Female Characters No Longer Get Additional Sewing and Cooking Skill

Optimism333:
## About The Pull Request
I don't believe there's any reason to hold on to this feature at this point. There is no longer a strength difference between the sexes (except in the case of dark elves), and female characters should simply get their skills based on the role chosen and any racial flavor modifiers (i.e. elves getting reading and dwarves getting mining).

Tested it locally and it worked fine.

## Why It's Good For The Game
- This no longer fits with the direction the server has gone, and there's no real reason to keep it. 

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/roguetown.dm
+++ b/code/modules/jobs/job_types/roguetown/roguetown.dm
@@ -54,9 +54,6 @@
 				H.patron = GLOB.patronlist[default_patron]
 		to_chat(H, "<span class='warning'>[ourpatron] had not endorsed my practices in my younger years. I've since grown acustomed to [H.patron].")
 	if(H.mind)
-		if(H.gender == FEMALE)
-			H.mind.adjust_skillrank(/datum/skill/craft/cooking, 1, TRUE)
-			H.mind.adjust_skillrank(/datum/skill/misc/sewing, 1, TRUE)
 		if(H.dna)
 			H.dna.species.random_underwear(H.gender)
 			if(H.dna.species)


NPC1314:
Consistent with the changes. Though truth is removing strength malus for women and race locks on jobs has made things less interesting and more slop.

Its questionable if the skill granter even works, it was broken at some point, but the pre_equip proc fix might have solved it. 
Then again, maybe not, and it doesn´t really matter I guess. Merge ahoy!

BadAtThisGame302:
Not a fan, don't see any reason why'd you remove it seeing as a lot of people that play woman get a nice bonus to their skills and I myself quite like it since not a lot of jobs get cooking or sewing.

Optimism333:
> Not a fan, don't see any reason why'd you remove it seeing as a lot of people that play woman get a nice bonus to their skills and I myself quite like it since not a lot of jobs get cooking or sewing.

I do not believe we are aiming for the sort of setting any longer where every single woman regardless of background and role will have experience in sewing and cooking.

A free skill bump solely based on the sex of the character seems something we should just move away from at this point.

Skills should be allocated based on the role you pick, and racial flavor bonuses. I think at this point that is a better direction for us.

ThePainkiller:
> A free skill bump solely based on the sex of the character seems something we should just move away from.
> Skills should be allocated based on the role you pick, and racial flavor bonuses. I think at this point that is a better direction for us.

### This. A million times this.

Except for **skills** being asigned by race, skills are and should be inherent to jobs. **Stats** however, should both be tied to race **and** job.

NPC1314 (approved):
Lets get this done.

PR #535, opened 2024-08-16
Paragive me para I have sinned

NPC1314:
Para some reason I suspect the new spiefling accent was not tested enough, or else we would have been parawarned so we would not paraget to stop this wordreplacement, a most unparatunate incident. Our mind is a paratress but only if we guard the gate, para obvious reasons. There is nont to fear however, this shall solve your troubles.

Other than this atrocity the accent is actually pretty good, the less word replacements the better generally. Less is more, somewhat exotic but legible.

## Pre-Merge Checklist

### DIFF
--- a/strings/spanish_replacement.json
+++ b/strings/spanish_replacement.json
@@ -5,7 +5,6 @@
 		"s": "ss",
 		"r": "rr",
 		"ish": "eesh",
-		"for": "para",
 		"is": "es",
 		"thanks": "gracias",
 		"in": "en"


SELFHELL:
Seems like I paragot about that, good change

ThePainkiller:
I've told people many times over for word replacers: NEVER DO THEM FOR THINGS BELOW 4 LETTERS

ThePainkiller (approved):

PR #537, opened 2024-08-16
craftable god helmets

YuiY1997:
## About The Pull Request

Adds templar helms that are already in the game to the crafting recipes

## Why It's Good For The Game

Drip
Not all races have access to templar without spending 20 triumphs
Impersonating holy people would be a fun gimmick
If people don't like that they're craftable they can regulate their creation IC

## Pre-Merge Checklist

itworks


Dorodomki:
i don't agree with the unique character gear being added to a crafting list, special roles get their unique special gear and it should be like that.

YuiY1997:
> i don't agree with the unique character gear being added to a crafting list, special roles get their unique special gear and it should be like that.

Paladin isn't a particularly special role. I also think more gear in general should be craftable. More drip > less drip

NPC1314:
Not the right way to go. I agree with Dorodomki on this one. Denied.

PR #541, opened 2024-08-18
Removes Aasimar from Captain and Hand

lavillastrangiato:
## About The Pull Request

See title.

## Why It's Good For The Game

Discussed with the dev team; we want to move towards aasimar being more discriminated against and seen more for their martial capabilities, and taking them out of roles that involve administration (thus why they remain in the royal guard, as they're the king's personal goons).

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/archivist.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/archivist.dm
@@ -12,7 +12,8 @@
 		"Humen",
 		"Elf",
 		"Half-Elf",
-		"Dwarf"
+		"Dwarf",
+		"Aasimar"
 	)
 	spells = list(/obj/effect/proc_holder/spell/invoked/projectile/fetch)
 	allowed_patrons = list(/datum/patron/divine/noc)


sanshoom:
It makes no sense to remove the /immortal/ race from potentially being an archivist.

Would you rather have a forty year old humen who hasn't seen anything of significance to be apart of your court as the archivist, or an immortal godling that has actually seen and experienced some of the large events that they themselves would have archived?

I can understand Captain and Hand, seeing as how the aasimar are known to get people killed due to their rash actions, but.. removing them from a noncombat role that they would totally be apart of doesn't make much sense. (Archivist isn't played much anyway, it's mostly for flavor, so I don't see any benefits from this.)

lavillastrangiato:
My original reasoning was because they're meant to be WAR-focused, and might not have/be considered sufficiently intellectual to be what is essentially a court scientist. However, I made a mistake, so I'll just make it captain and hand right now. :)

DovaKitty (approved):
This was discussed in the staff meeting as a good idea. While not explicitly oppressed like Tieflings, it makes sense that the very rare Aasimar would not be in these positions of authority.

NPC1314 (approved):
When checked for TM conflicts, mergetime

PR #538, opened 2024-08-17
Garrison Rebalancing (Skills)

Optimism333:
## About The Pull Request
In line with the previous balance prs, brushes up the skills of various garrison roles. Garrison guards were falling a bit behind some of the basic combat classes, even with the mercenary rebalancing in #533 , so I buffed them a little bit, along with royal guards. I made garrison archer's melee skill consistent with the mercenary changes, and I added a few skills to Captain that were missing or lacking.

### Garrison Guard ###
- Garrison Guards no longer get +1 to Perception
- Garrison Guards get skilled (3) athletics instead of average (2)
- Garrison Guards get skilled (3) swords instead of average (2)
- Garrison Guards get skilled (3) shields instead of average (2)

### Garrison Archer ###
- Garrison Archer has had expert (4) knife reduced to skilled (3)
- Garrison Archer has had skilled (3) wrestling reduced to average (2)
- Garrison Archer has had tanning and crafting skills set to weak (1) instead of a chance of weak (1) or average (2)
- Garrison Archer has had bow skill reduced from master (5) to expert (4)
- Garrison Archer has been given average (2) swords, instead of a chance of none (0) or weak (1)

### Royal Guard ###
- Royal Guards get skilled (3) knives instead of average (2)
- Royal Guards get expert (4) athletics instead of average (2)

### Captain ###
- Captain starts with a metal shield, like Royal Guards
- Captain has expert (4) shield skill, previously had none
- Captain has expert (4) axes/maces, instead of skilled (3)
- Captain has skilled (3) bows and crossbows, previously only having average (2) crossbow skill
- Captain has higher climbing, swimming, and riding skill. Previously he was worse at all of these than a Royal Guard

### Misc Other ###
- Veteran has skilled (3) athletics instead of average (2)
- Dungeoneer has average (2) athletics, previously had none
- Squire has average (2) athletics instead of weak (1)

## Why It's Good For The Game
- Brushes up a lot of garrison stats, bringing them up to par with recently pr'd balance changes for other combat classes

## Pre-Merge Checklist

- [x] You tested this on a local server.
- [x] This code did not runtime during testing.
- [x] You documented all of your changes.

PR #544, opened 2024-08-18
The LEVYING! (Garrison Class Selection - As Promised)

Stutternov:
## About The Pull Request

Basically adds a garrison 'class' selection system, this way we don't have to have garrison archer and garrison split into separate roles; as well as allowing specialization loadouts for people to choose a sort of 'class' that they specifically like. I've seen often guards be a bit upset they roll weird and can't use their favorite playstyle. So, here, this should alleviate that! 

This acts like an Adventurer, giving you the round pop-up 'select your class' list but of only guard classes! Yippie!

### The Guard Types

Four guard classes have been added as selectable under the garrison job-type. They are as follows:

- **Footsman**, the most basic of guard types. Footsmen basically use what guards prior had with some small changes, focusing on axes, maces, and shields.
- **Pikesman**, a guard with slightly more strength by one but less speed by one in turn. Pikesmen start with either a spear or billhook and sport a slightly different look via their gear.
- **Archer**, unchanged from garrison archer but merged into the role. Exact same as before with the RnG removed; now you roll 1 crafting, 1 tanning, 1 sword. (As opposed to: 1-2 crafting, 1-2 tanning, 0-1 sword. Standardized.)
- **Arcane Enforcer**, a guard focusing on _magic_ usage, mixing together a few combat stats and lesser non-lethal spell types. Limited to 2 slots these guards are still push-overs compared to others in combat but serve a massive support use, and can hold themselves to a minor extent; though are out-preformed by nearly any combat-dedicated job.

## Why It's Good For The Game

Adds variety to the Garrison roles. This type of PR has been done on other servers with pretty positive reception. It allows people to play into specific types of roles they like and act as a character with specialties rather than expecting to be well-rounded.

This was requested by staff and some players, and I don't see why not. My only need is feedback; as roles, such as the Arcane role, is a bit up in the air as its balance is hard to determine without actual playtesting with players vs players. While they are actually weaker spells-wise than any other mage role, they do have some bonus stats (mostly in speed) over those roles, so the balancing might be alright..?

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/garrisonguard.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/garrisonguard.dm
@@ -166,7 +166,7 @@ Design philosphy:
 	H.mind.adjust_skillrank(/datum/skill/misc/climbing, 3, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/misc/athletics, 3, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/misc/lockpicking, 2, TRUE)
-	H.mind.adjust_skillrank(/datum/skill/combat/swords, 1, TRUE) // For rapier use
+	H.mind.adjust_skillrank(/datum/skill/combat/swords, 2, TRUE) // For rapier use
 	H.mind.adjust_skillrank(/datum/skill/craft/crafting, 1, TRUE)
 	H.mind.adjust_skillrank(/datum/skill/craft/tanning, 1, TRUE)
 	H.change_stat("intelligence", 1)


Stutternov:
Note: This is drafted while I receive feedback from the staff members who requested this, coder-staff for balance helping, and community feedback on the general idea.

sanshoom:
call arcane enforcers warlocks :wise_tree:

code/modules/jobs/job_types/roguetown/garrison/garrisonguard.dm
```diff
@@ -41,42 +43,181 @@
...
+
+	category_tags = list(CTAG_GARRISON)
+
+/datum/outfit/job/roguetown/guardsman/pikesman/pre_equip(mob/living/carbon/human/H)
+	..()
+	//Armor for class
+	armor = /obj/item/clothing/suit/roguetown/armor/chainmail
+	neck = /obj/item/clothing/neck/roguetown/gorget
+	head = /obj/item/clothing/head/roguetown/helmet/kettle
+	if(prob(50))
```
ThePainkiller, line 116: Lower prob please, like 33% or something. Billhook is a really good weapon, and a good justification for them to purchase one from a smith.

code/modules/jobs/job_types/roguetown/garrison/garrisonguard.dm
```diff
@@ -41,42 +43,181 @@
...
+	H.mind.adjust_skillrank(/datum/skill/combat/bows, 5, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/combat/crossbows, 4, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 3, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/combat/unarmed, 2, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/misc/swimming, 3, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/misc/reading, 1, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/misc/climbing, 3, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/misc/athletics, 3, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/misc/lockpicking, 2, TRUE)
+	H.mind.adjust_skillrank(/datum/skill/combat/swords, 1, TRUE) // For rapier use
```
ThePainkiller, line 169:
```suggestion
	H.mind.adjust_skillrank(/datum/skill/combat/swords, 2, TRUE) // For rapier use
```
Considering last admin meeting with a consensus to give non-combat weapons a standardized level 2 on weapon skills, up this to swords 2 for consistency.

ThePainkiller (changes requested):

NPC1314:
Good concept.
Skills should be aligned with #538
Perhaps add a root pre_equip for the skills in common to make it easier to overview?
Mage guard is not a good idea IMO, should be kept to court mage, occasional wandering mage, inflation
of number of characters with the same special talents makes them a lot less interesting, also balance problem.

1N7Slayer:
I like the arcane enforcer idea. Roguetown has been too magic-lite for awhile, so I'd like to see this server stand out from the others by comparison.

code/modules/jobs/job_types/roguetown/garrison/garrisonguard.dm
```diff
@@ -41,42 +43,181 @@
...
+	..()
+	//Armor for class
+	armor = /obj/item/clothing/suit/roguetown/armor/leather/hide
+	beltr = /obj/item/reagent_containers/glass/bottle/rogue/manapot		//Fortnite chugjug
+	backr = /obj/item/rogueweapon/woodstaff								//Stinky-uggy wizard staff. Go do kung-fu.
+	head = /obj/item/clothing/head/roguetown/helmet/kettle
+	neck = /obj/item/clothing/neck/roguetown/coif						//Screw you, magic-man. You get a worse coif!
+
+	//Stats for class
+	H.mind.adjust_skillrank(/datum/skill/combat/knives, 3, TRUE)
```
Valkrae, line 198: personally, i think any weapon skill above 3 might be too much.

ThePainkiller: But this is exactly 3, as expected of every town guard as per our admin meeting...

Valkrae: its a mage, though. not a direct combat class. it's supposed to be a support element, that's why I'm saying any weaponskill above 3. Sorry, it doesn't show it in the review, but this particular 3 knives is for the Arcane Enforcer class.

Mojamble:
I understand why people don't like the idea of guard mages, but at the same time, is a taser guard REALLY such a bad idea? Would lead to more non-lethal take downs, plus they'll get crumpled in a real fight.

Optimism333:
## Review ##
### Footman ###
image

- Would suggest giving this class skilled (3) shields at minimum, and also skilled (3) swords, so they can use their board with a sword, axe, or mace depending on preference.
- I don't think they need polearm skill (since that is the focus of the Pikeman class), or riding skill
- I would also suggest slightly adjusting their stats to something like this: +1 strength, +2 endurance, +1 constitution. Dropping that point of perception to give them a bit more stamina for those drawn out close quarters fights

### Pikeman ###
image

- Pikeman stats are probably alright as is, would just remove the +1 to perception as they do not need it
- I would suggest giving them skilled (3) in polearms and axes/maces (to use the grand mace, and potentially the long axes that will come with Dromkii's prs), but no shield skill

### Garrison Archer ###
image

- Would suggest dropping knife skill from expert (4) to skilled (3), to match with the baseline for common combat classes. Other than that, this one seems fine

### Garrison Arcane Enforcer ###
image

- I would suggest just giving them skilled (3) polearms so they can use their staff for wizard-fu, and maybe average (2) knives
- If they're apprentice level magic users, I would reduce their Intelligence to +2, and perhaps reduce their -2 in strength to -1
- This class is a bit of a balancing eniga, to be frank. But hopefully we can figure out what fits best in time

DovaKitty (changes requested):
These lines of code are not implemented in garrisonguard.dm and will need to be added for subclass selection to function as mercenary does. Currently this PR spawns guards in with just their base gear with no subclass select.
image

Stutternov:
Note: This PR is waiting for nets to go in fully. I have an idea......

ThePainkiller:
> Note: This PR is waiting for nets to go in fully. I have an idea......

Nets are fullmerged, nerd. Go ham.

NPC1314:
SK divorded 544 and married the younger version in 809. Time to get old yellered.

PR #547, opened 2024-08-18
Stops Astrata from torching vampires inside of their mining cave

bowlofcereal:
## About The Pull Request

Changes the area from under/cave to indoors for the ore cave in the vampire-manor on CentCom, because for some reason this doesn't count as underground like the rest of the building. I'm fairly certain this is because there's only 1 Z-Level layered here, so there is technically no roof for the cave.

This has been localtested and it works fine.

## Why It's Good For The Game

It's a little immersion breaking to have the mining cave somehow outside and exposed to the elements, and i'm also tired of watching vampires, or worse yet the VL, walk in there thinking it's perfectly safe only to combust.

## Pre-Merge Checklist


ThePainkiller:
To be honest, rid vampires of their own mining cave has been an idea pitched and planned for for a long time, and a preferrable alternative to this.
They shouldn't get their own town2 with a great forge and all that, with how OP they currently are.

NPC1314:
> To be honest, rid vampires of their own mining cave has been an idea pitched and planned for for a long time, and a preferrable alternative to this. They shouldn't get their own town2 with a great forge and all that, with how OP they currently are.

Agreed but better to fix this and do the VL fix proper instead of leaving in the janky area. This PR obsolete with PR #548, closing this later unless theres a reason to keep it open?

bowlofcereal:
> > To be honest, rid vampires of their own mining cave has been an idea pitched and planned for for a long time, and a preferrable alternative to this. They shouldn't get their own town2 with a great forge and all that, with how OP they currently are.
> 
> Agreed but better to fix this and do the VL fix proper instead of leaving in the janky area. This PR obsolete with PR #548, closing this later unless theres a reason to keep it open?

Nope-- completely fair if you're merging it into something else. Closing on request.

PR #540, opened 2024-08-17
Temporarily makes bandits a drifter class, plus gives them 'backgrounds'.

Marmio64:
## About The Pull Request

Bandits have been broken for a long time and I am not a good enough coder to fix them the right way. However, I am able to present a temporary alternative way of allowing them to still function by making them simply be a drifter class.

This drifter class will just turn them into a bandit, putting them in the bandit camp and giving them all the skills and what not. The class itself has a 100% pick rate/probability of being available, and has a max of 5 slots.

This also gives bandits the ability to choose their 'background' rather than them just having slightly randomized gear/stats. The backgrounds do not get expert/master skills in a weapon or expert wrestling/unarmed (unless poacher with bows/crossbows) like bandits used to have, but to compensate they have better gear and/or traits. In addition, bandits have average swimming, lockpicking, and stealing skills now, and their intelligence malus is -2 instead of -4.

For now there are three available bandit backgrounds, of which they are pretty comparable with standard drifter roles-

Deserters: Former soldiers who committed the crime of desertion. They start with pretty good armor with a rusted barbute (heavy helmet, but limits view and has mediocre durability), iron coif, rusted half-plate (same as normal, but halved max integrity), and iron chausses, along with leather bracers, gloves, and boots. In addition they have skilled polearms and have a spear with them. Most importantly, they have medium armor training.

Poachers: Hunters who operate illegally. They are speedy, have good perception, and get the dodge expert trait. They also have great bow and crossbow skills, and are good with knives. Though, they do have lower strength than the other backgrounds, and less armor in general, as they only have a gambeson, volf head (weaker leather helmet), leather gloves and trousers, and boots.

Brigands: Full-blooded bandits who have been in the business for a while. They are sort of a jack of all trades and have a good stat line and are skilled with flails, swords, shields, clubs, and crossbows. They also get a decent level of protection with leather
armor, leather coif, leather trousers, leather bracers, leather gloves, and leather boots. While good at many things, they don't get unique traits like the other backgrounds.

also adds some special armors for them, with sprites done by cre. From left to right: Brigand, Poacher, Deserter.
image

This will also make it so examining wanted posters will actually let you discover bandits thereafter.

This will definitely need testing. It seems okay right now, nothing game breaking at least, but you never know.

## Why It's Good For The Game

Gets bandits back into the game. Not in a optimal way, but I feel like having them would be better than not.

Also gives them some neat classes to choose from.

[...]

PR #548, opened 2024-08-19
Centcomm

NPC1314:
## About The Pull Request

Adds the lair, includes the area change in #547 and incorporates bits of #540 so we can just do this map change in one go and not drag it out

## Pre-Merge Checklist


ThePainkiller (approved):
The bathroom stall killed me

PR #559, opened 2024-08-22
Nukes Weaving

SELFHELL:
## About The Pull Request
It REALLY shouldnt have been a separate skill to start. I made a mistake on that bit

Also fixes nun clothing, apparently it was broken?

## Why It's Good For The Game

Skill bloat bad or something, sewing/weaving do functionally the same thing

## Pre-Merge Checklist


ThePainkiller (approved):
tbh yeah, skillbloat was real, just clothes crafting is fine in general, no need to divide using silks or using fibers when it's all the same, craftdiff exists for a reason after all

PR #563, opened 2024-08-23
Projectile accuracy (Your target zone is not guaranteed)

Marmio64:
## About The Pull Request

Adds a new variable to projectiles known as 'Accuracy', which determines the chance you will hit your intended target zone.
If you fail the accuracy check, you still hit your target if you were supposed to. However, the zone you hit them is will be random rather than your intended zone, and is weighted towards hitting the enemy's chest.

Accuracy is by default 75, meaning a 75% chance to hit your intended target at point blank. This increases by 5 per perception point above 8, meaning you need 14 perception to 100% of the time hit a precise target at point blank.

Accuracy will decrease by 10 per travelled tile, to a cap of 11 accuracy.

In addition, fully charging/aiming your shot adds a flat +20 accuracy buff. (If you fully charge a shot and have at least 9 perception, you will always hit your intended target at point blank)

Generally, without looking ahead, your target should be about 5 tiles away if you are maintaining  distance. With good perception you should have a 50/50 chance on hitting your specific target. If you fully charge/aim your shot, you will increase your chance of hitting your specific target to about 70%.

## Why It's Good For The Game

Ranged combat is incredibly powerful against anyone who leaves even a small bit of themselves not protected by plate, as you are guaranteed to hit your target zone and just one or two arrows are often enough to kill someone by bleeding them out.

This will slightly decrease how powerful it is, since like melee, you will no longer be guaranteed to hit a specific body part.

## Pre-Merge Checklist


Catalaria:
Dude... this is so based... so zased. Finally perception meaning something.

ThePainkiller:
Can we please have estimate accuracy on distance? I feel like the dropoff is insane, and that it shouldn't apply to crossbows to give them an advantage (and because it makes sense since they have greater pull strength)

Cre77:
you could include the actual bow and crossbow skills into the equation, seeing as how otherwise those skills would only ever affect drawing/aiming speeds

ParadoxalObserver:
> Base arrow accuracy reduced to 65, and base bolt accuracy to 85. Fully aiming also gives +15 accuracy instead of +20.
> 
> However, each skill level in bows will increase accuracy by +5, and each skill level in crossbows give +3. (Crossbows are easier for lay people to use, however there is less aiming to master.)
> 
> Having a skilled level in bows will for the most part be the same as it was before this change, same thing for crossbows. However, if you're not very good your accuracy will be slightly worse, while if you're great your accuracy will be better.

While I think this is an excellent PR, I feel the numbers are a bit whacky. For one, the numbers sort of feel like they fall under the assumption that a person is hitting every single one of their arrows, which isn't the case. The thought that I have twenty arrows and only 1-in-10 will hit the zone I want if I'm shooting supportively (in an event, for example) is... questionable. The change I'd make is that each skill point increases the minimum by 5. So at Archery 3, your minimum is 26. It's a 1-in-4 chance and this is assuming you actually land your shot to begin with, which can be easier said than done.

NPC1314:
Good concept, getting a testmerge to see how it holds up in practice

NPC1314 (approved):

PR #564, opened 2024-08-23
Matthios idol loot tweaks

Marmio64:
## About The Pull Request

The idol now gives you more applicable loot.

At the first tier of 'rewards', you are only given scale mail if you have the training to actually wear it. Otherwise, you get iron chainmaille.

The second tier is kept the same, you always get the horned helmet.

The third tier, the weapon you get is dependent on your skills. Deserters will get the standard bill hook, poachers will instead get a long bow, and brigands get a steel longsword.

Adds a fourth tier of loot, which will give you steel chain chausses for leg protection.

In addition, the idol will now accept jewelry such as circlets and rings, alongside fineries such as gold/silver goblets and statues. However, these fineries (alongside gems) will only be worth half their actual sell value, to encourage fencing them first for coins if you're able to do so. If you can't you will just have to give them to Matthios at a reduced value.

## Why It's Good For The Game

Makes the idol give universally useful loot for those who give offerings to it, rather than the loot mainly being geared towards only being useful for the deserter background.

Also generally makes the loot be better and more easily achievable as you can now sell a lot more things to your god.

## Pre-Merge Checklist

### DIFF
--- a/code/game/objects/structures/fluff.dm
+++ b/code/game/objects/structures/fluff.dm
@@ -1014,6 +1014,8 @@
 					B.contrib -= 100
 					var/obj/item/I
 					switch(B.tri_amt)
+						if(1)
+							I = new /obj/item/reagent_containers/glass/bottle/rogue/healthpot(user.loc)
 						if(2)
 							if(HAS_TRAIT(user, TRAIT_MEDIUMARMOR))
 								I = new /obj/item/clothing/suit/roguetown/armor/plate/scale(user.loc)


NPC1314:
Blackstone/Ratstone added some more levels/variety to their loot list, like first level is something like a health potion, moondust, then you get armor. Has utility and you still might want to try to loot/rob armor pieces for a bit longer and get more diverse gear that way.

Something like that as first tier, with some important armor accessory (neck/legs I guess) as second, then armor/helmets seems like it might be a better scale, slightly weighting the chance for more diversity in armor compared to having tier I reward be chainmail or scale, then all bandits will have one of those almost every round.

Can you tweak it a bit?

NPC1314 (approved):

PR #566, opened 2024-08-23
RNG Stat variation now weighted towards unchanged, rather than deducted

saltprjak:
## About The Pull Request

Changes the way random stat variation is handled to now favor stats remaining as-is, with a 1 in 3 chance for any stat to be different from baseline in either direction. Before that, it was weighted towards stats receiving a -1, which results in everyone having less stats across the board.

## Why It's Good For The Game

In terms of PvP - this doesn't change anything either way, but this does bring crafting and training in line with intended pace, and helps maintain general sanity in balancing discussions now that all characters aren't generally weaker than they are intended to be.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/stats.dm
+++ b/code/modules/mob/living/stats.dm
@@ -35,7 +35,7 @@
 	STASPD = 10
 	STALUC = 10
 	for(var/S in MOBSTATS)
-		if(prob(50))
+		if(prob(33))
 			switch(pick(1,2))
 				if(1)
 					change_stat(S, 1)


Optimism333:
Good stuff. The math on this was always odd to me.

sanshoom:
holy peak

NPC1314 (approved):
Why one would design the system so that the norm is -1 instead of 0 makes no sense to me, 0 is the very definition of the middle ground, the norm, the baseline.

Doing TM to be on the safe side

PR #571, opened 2024-08-23
Feint now works against dodge intent.

1N7Slayer:
## About The Pull Request

You can now feint someone dodging.

## Why It's Good For The Game

Dodge is sort of zero interactivity in close combat. The fact it sets your chance directly to zero, no modifiers or anything like that, is pretty skewed towards dodge-based classes such as assassin or rogue in CQC. This aims to level the parity between fast and slow classes to also include 'smart' classes to be able to beat fast classes in combat.

## Pre-Merge Checklist

### DIFF
--- a/code/game/objects/items/rogueweapons/rmb_intents.dm
+++ b/code/game/objects/items/rogueweapons/rmb_intents.dm
@@ -33,8 +33,6 @@
 				perc += (ourskill - theirskill)*10
 	if(user.STAINT < L.STAINT)
 		perc -= 15
-	if(L.d_intent == INTENT_DODGE)
-		perc = 0
 	if(!L.cmode)
 		perc = 0
 	if(L.has_status_effect(/datum/status_effect/debuff/feinted))


ParadoxalObserver:
So, this is a bad change, here's why:
-Classes that rely primarily on dodge intent tend to have low strength and con, making them susceptible to stuff like sprint charges.
-Dodge is purely SPD vs. SPD, so even if you're a class that utilizes dodge, your benefit is very easily nullified if you're fighting an elf, goblin, vulf, really anything with high SPD.
-Dodge gets a multitude of penalties from pain, if you get hit once it's typically game over.
-Utilizing armour heavier than light with dodge will kill your stamina.
-Short weapons give the attacker a higher multiplier on their tohit against dodge.

tl;dr get good

A bit longer: if you are playing a character that can actually utilize dodge, then you have a variety of weaknesses that make you very, very easy to defeat. Dodge has received countless nerfs despite it not being good and anyone with the bare minimum knowledge of combat code knows how to beat it.

Dodge is already weak to strength and con focused characters, why on Earth are you adding in a weakness to int and skill?

NPC1314:
ParadoxalObservers kind of delivered a killing blow there, it's game over.

PR #579, opened 2024-08-25
Ports Ratwood's "Charge and Sprint changes"

saltprjak:
## About The Pull Request

*now with minor changes
Ports https://github.com/Rotwood-Vale/Ratwood-Keep/pull/475. Tl;dr - Charging now takes into account distance travelled, and a combination of STR and CON for the attacker, STR and END for the defender, and athletics for both. Running only few tiles makes it harder to knock someone over, running a longer distance and/or from behind makes it easier. 
Tested, works as intended.
## Why It's Good For The Game

Removes the ability to easily stat-check someone by charging into them point-blank and stealing their weapon. Allows weaker characters to punch above their weight in that regard with good positioning. Also makes gobbos and skeletons much harder to cheese for the same reasons.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/admin/verbs/schizohelp.dm
+++ b/code/modules/admin/verbs/schizohelp.dm
@@ -15,6 +15,7 @@ GLOBAL_LIST_EMPTY_TYPED(schizohelps, /datum/schizohelp)
 	var/display_name = get_schizo_name()
 	var/message = span_info("<i>[display_name] meditates...</i>\n[msg]")
 	var/message_admins = span_info("<i>[display_name] ([key || "NO KEY"]) [ADMIN_FLW(src)] [ADMIN_SM(src)] meditates...</i>\n[msg]")
+	log_game("([key || "NO KEY"]) mentorhelped: [msg]")
 	for(var/client/voice in (GLOB.clients - client))
 		if(!(voice.prefs.toggles & SCHIZO_VOICE) || check_rights_for(voice, R_ADMIN))
 			continue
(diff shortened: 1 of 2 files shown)


CormanQuewye:
I can't believe this was done at the same time as I was working on my own sprint changes. I like it accounting for distance travelled, but I do dislike the calculation *again* relying purely on STR and CON with zero effect from athletics and not being weighted towards the defender, effectively not really fixing the issue of this being an easy-to-employ strategy for anyone with higher con as opposed to a high-risk and high-reward strategy primarily meant to be employed by undisciplined fighters.

Cre77:
after discussing with Corman, defenders now use END instead of CON (since endurance is way more flexible with buffs and mood, and harder to gauge for sweats), athletics now affects the calculation, there's tf2 random crits, and the distance-needed-to-travel punish is slightly harsher

NPC1314:
So how does this mesh with #608 ? 

Havent heard any feedback on this one, not sure where we at.

CormanQuewye:
> So how does this mesh with #608 ?
> 
> Havent heard any feedback on this one, not sure where we at.

#608 were extra mechanical additions I had in #585 since the maintainers looked to prefer these sprint changes to mine (and tbf this one has some neat stuff mine didn't like modifiers depending on distance ran) I ended up closing that one and moving those extra tidbits people seemed to like into its own PR, which is #608

tl;dr nothing really, #608 is just a few more changes relating to knockdowns which don't conflict with this sprint PR but add a little more to combatant through a few simple changes

ThePainkiller (approved):
Considering this PR has been up for a while and I've seen nothing but people confused they can't tackle their way out to victory, I consider this PR an absolute win. Good for fullmerge.

PR #594, opened 2024-08-27
Hand nerf

DovaKitty:
## About The Pull Request

* Brings down Hand's combat skills. Previously they had 4 swords skill, now they have 3 knives skill. They spawn with a knife and are spymasters, so it is fitting.
* Increases lockpicking skill to 3 from 2. Any doors they cannot open with their keys, they can do so this way.
* Reduces armor training from heavy to medium.
* Replaces Hand's +2 strength with +1 speed.

## Why It's Good For The Game

This PR adjusts Hands from being renamed Knights to something more fitting of their role as a master of court intrigue. They can still best non combat classes, but will struggle fighting the dedicated fighter classes like royal guard. There is really no excuse for how strong they were previously, considering the amount of power they had and the 120 mammons they spawned with. Powergamer class no-more.

## Pre-Merge Checklist


BadAtThisGame302:
I love nerfing towner roles so much.

ThePainkiller:
Still iffy they get armor training, at all

ThePainkiller:
Yeah, upon further revision, if they get a point of speed, they'd be better off with dodge expert rather than medium armor training, it fits the whole "spy" vibe better.

DovaKitty:
> Yeah, upon further revision, if they get a point of speed, they'd be better off with dodge expert rather than medium armor training, it fits the whole "spy" vibe better.

Done, swapped out armor training for dodge. Makes elf Hand better but other races worse. I don't really agree with it, but it does fit them more.

ThePainkiller (approved):
They are still pretty cracked with daggers 3 and swords 2 tbh, crossbows is an even extra, but for now it still is a nerf to how insanely good they were.

PR #596, opened 2024-08-27
move aside, poor weak ale, make room for beer

cornercube:
## About The Pull Request
adds a more proper beer to the game
regular recipe produces an ale, which is fermented without the use of hops

however, you can now harvest some bog-bitter from reeds in the bog
which can be combined with the grain to make a wort
fermenting the wort produces a nice, dark, stout
booze power and flavor is higher

also adds some some quality levels to importable beers

## Why It's Good For The Game
adds some flavour (HA)

## Pre-Merge Checklist

### DIFF
--- a/modular/Neu_Farming/code/produce.dm
+++ b/modular/Neu_Farming/code/produce.dm
@@ -265,7 +265,7 @@
 	eat_effect = /datum/status_effect/debuff/badmeal
 	list_reagents = list(/datum/reagent/drug/nicotine = 5, /datum/reagent/consumable/nutriment = 1)
 	grind_results = list(/datum/reagent/drug/nicotine = 10)
-  possible_potion = "poison"
+ 	possible_potion = "poison"
 	dust_result = /obj/item/alch/tobaccodust
 
 // -------------- HOPS and BEER WORT -----------------


code/modules/reagents/chemistry/reagents/alcohol_reagents.dm
```diff
@@ -72,7 +72,7 @@ All effects don't start immediately, but rather get worse over time; the rate is
 	return ..()
 
 /datum/reagent/consumable/ethanol/beer
-	name = "Beer"
+	name = "Ale"
```
ThePainkiller, line 75: Beer is a catch-all term for ale, technically all variants on this file are _types_ of beer, Ale is one of the two classifications for beer (Lager/Ale, which depends largely on the temperature it was brewed)
Seeing as how this is the base, why not make the poor man's choice just remain Beer, and make Ale into another type with a different recipe?
Also, you can't turn Ale into a stout since a stout is a lager, and not ale (the difference is stupid but...cohesion)

cornercube: the basic beer is supposed to be ale, since it has no hops
adding hops to it is supposed to make it into more of a regular beer

PR #597, opened 2024-08-27
Cleans up the last of the mage revert

NPC1314:
## About The Pull Request

With the return to focusing on court mage being the assigned sumbucket the ring of protection is not needed anymore.
(In fact should have been reverted with the hat stuff but was overlooked. With this the revert should be complete since removing the debugged fireball and spell art would be silly.)

The ring also restored to full function and the too long names of the guild rings shortened a little so they fit with the rest in inventory displays.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/magician.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/magician.dm
@@ -35,7 +35,6 @@
 	id = /obj/item/clothing/ring/gold
 	belt = /obj/item/storage/belt/rogue/leather/plaquesilver
 	beltr = /obj/item/keyring/mage
-	id = /obj/item/clothing/ring/gold/protection
 	backl = /obj/item/rogueweapon/woodstaff
 	shoes = /obj/item/clothing/shoes/roguetown/shortboots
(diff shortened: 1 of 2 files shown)


code/modules/jobs/job_types/roguetown/nobility/magician.dm
```diff
@@ -35,7 +35,6 @@
 	id = /obj/item/clothing/ring/gold
 	belt = /obj/item/storage/belt/rogue/leather/plaquesilver
 	beltr = /obj/item/keyring/mage
-	id = /obj/item/clothing/ring/gold/protection
```
BadAtThisGame302, line 38: Give the Mage an anti-magic ring then to compensate.

NPC1314: No. They literally got it to compensate for having age bracketed spells and the hat, with that gone, so goes the ring.

ThePainkiller: If the ring is gone from them, is it even obtainable?

Stutternov (approved):

PR #599, opened 2024-08-27
Tell Good Jackberries From Bad

lavillastrangiato:
## About The Pull Request

With a Farming skill of 3 or above, you can now discern poisonous jackberries apart from good ones by examining them.
image

@Optimism333 and @DovaKitty coded this and helped input it at my request! I am merely testing and PRing.

## Why It's Good For The Game

Makes Farming skill more useful, and allows a way to discern poisonous berries without just eating and praying.

## Pre-Merge Checklist


CormanQuewye:
A decent change! Funnily enough, you can already do this if you have an arrow by checking if the poison arrow craft pops up (which makes the thumbs down someone gave this extra funny).

Optimism333:
Don't understand the thumbs down, myself. I think it's a great change, even if it will likely only be relevant at the start of the round. But if we were expand foraging with new items, this check could easily be be expanded.

Heck, it would be neat if high intelligence and/or perception also allowed the check to be made.

Maybe around the range of 14+ in either category. I might run some tests with code and see if that can be applied.

CormanQuewye:
> Don't understand the thumbs down, myself. I think it's a great change, even if it will likely only be relevant at the start of the round. But if we were to expand foraging with new items, this check could easily be be expanded.
> 
> Heck, it would be neat if high intelligence and/or perception also allowed the check to be made.
> 
> Maybe around the range of 14 or above in either category. I might run some tests with code and see if that can be applied.

You could use d20 ttrpg logic and apply a +1 modifier to the skill check for every 2 INT above 10.

NPC1314 (approved):

PR #604, opened 2024-08-28
VL mode pop limit port

NPC1314:
## About The Pull Request

Copied quick and dirty solution from Ratwood. Supposed to stop VL mode from occuring when less than 15 people readied up in lobby (can set any number I guess)
Ports: https://github.com/Rotwood-Vale/Ratwood-Keep/pull/851/files

Not able to test this offline, don't know how well it actually works, I dunno.

## Pre-Merge Checklist


NPC1314:
Lets do a testmerge and see how it goes

1N7Slayer:
15 players readied is a bit too much, lower it down to 10.

Stutternov (approved):

PR #608, opened 2024-08-28
Combatant Butterfinger Remover (now with 20% more tripping)

CormanQuewye:
## About The Pull Request

My former sprint tackle PR with the sprint-tackling removed (ironic) but the little additions still in. Done so people can focus their sprinting discussion on the actual sprint PR.

What does this add?
- Knockdowns do not cause you to drop held items, as long as you have combat mode on (unsuspecting people still will);
- Sprinting over corpses or people who are lying down will cause you to trip;
- Adds a distinction between knockdown and paralyze in the variables viewer.

## Why It's Good For The Game

Knockdowns currently are incredibly strong to the point that pulling one off is typically a gameover for your opponent. This has lead to an over fixation on STR and CON and methods of getting knockdowns as a way to get through every fight. If you are lying down you already receive maluses to defense, are an easy target, and have a heavily restricted range to hit on the target doll. The loss of your weapon is both silly and essentially gives the opponent multiple free hits on top of all the other benefits.

This'll lead to more consideration on if putting the effort into seeking a knockdown is worth it considering the situation and will lead to wrestlers considering other avenues beyond "grab chest > tackle" as now their opponent will still be capable of stabbing them repeatedly with that dagger they chose to ignore.

The tripping on lying down mobs is meant to both reward an active defense against sprint charges (if you quickly lie down you'll have a brief window of advantage) and to discentivise sprint and charging on body-laden battlefields (this is mostly for events).

And a little addition to the var viewer that makes a distinction between knockdown and paralyze, as a treat.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -782,7 +782,13 @@
 	for(var/i in get_equipped_items())
 		var/obj/item/item = i
 		SEND_SIGNAL(item, COMSIG_ITEM_WEARERCROSSED, AM, src)
-
+	if(isliving(AM))
+		var/mob/living/L = AM
+		if(L.m_intent == MOVE_INTENT_RUN && lying && !buckle_lying)
+			L.visible_message("<span class='warning'>[L] trips over [src]!</span>","<span class='warning'>I trip over [src]!</span>")
+			L.Knockdown(10)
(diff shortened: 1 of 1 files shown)


Stutternov (approved):
Based and trip-pilled.

PR #609, opened 2024-08-28
Moves the bonus wielding Wdefense from 3 to 1

saltprjak:
## About The Pull Request

hold the salt.
Wielding a weapon now gives only 1 bonus wdefense, compared to the prior 3.
Kite shield's wdefense from 8 to 7 (to follow the others' natural progression of 5 and 6). Aasimar shield is still at 8.

## Why It's Good For The Game

Right now, when it comes to standard melee, wielding a two-handed weapon is the better option for most combat roles on the server. Wielding a weapon in two hands gives you the benefit of almost doubling its defense value, which resulted in two things:
1) You can easily reach 80-99% parry chance with even the worst weapons, resulting in simplemobs and unskilled NPCs becoming complete pushovers so long as you had 2-3 weapon skill (which, with recent updates, is something almost everyone has now).
2) Outside of arrow protection and SPD-diff swift weapons, this left shields, one-handed weapons and the option of leaving your two-hand weapon in one hand in favour of something else in the dust coupled with all the utility alt intents and massively improved damage.

This will, for example, put wielding swords (4 wdefense + 1 from wielded) on par with the bottom-tier wooden shields (5 wdefense), and make high-end weapons like halberds (5 wdefense + 1 from wielded) slightly worse than the highest-tier kite shields (7 wdefense - this also nerfs them to prevent powercreep, and make the progression smoother).
In essence, this just means the high damage from wielded weapons will come at a greater tradeoff, and combat in most cases will become deadlier now that this basically shaves off 20% parry chance from the game's wielded arsenal.

**This is by no means a complete rehaul of the combat system, merely a small first step towards knocking the longsword and the halberd off their undisputed meta king throne.** Memey stuff like dodge and swift weapons can be addressed in other PRs, especially since people using them are a small minority compared to the rest of the game's melee fighters.

Wdefense can still reach pretty silly numbers once you take smithing quality wdefense bonuses into account, and reaching 99% against unskilled opponents is still easy for dedicated combat role basekits.

## Pre-Merge Checklist

### DIFF
--- a/code/game/objects/items/rogueweapons/shields.dm
+++ b/code/game/objects/items/rogueweapons/shields.dm
@@ -168,7 +168,7 @@
 	wlength = WLENGTH_NORMAL
 	resistance_flags = null
 	flags_1 = CONDUCT_1
-	wdefense = 8
+	wdefense = 7
 	coverage = 70
 	attacked_sound = list('sound/combat/parry/shield/metalshield (1).ogg','sound/combat/parry/shield/metalshield (2).ogg','sound/combat/parry/shield/metalshield (3).ogg')
 	parrysound = list('sound/combat/parry/shield/metalshield (1).ogg','sound/combat/parry/shield/metalshield (2).ogg','sound/combat/parry/shield/metalshield (3).ogg')


Optimism333:
I think this is a step in the right direction to make shields a more attractive option

NPC1314 (approved):

PR #614, opened 2024-08-29
Knight Crests

WWWWW-Git:
## About The Pull Request
Adds four new helmets that can be crafted by smithing. These helmets are resprites of the knight, hounskull, bucket and bucket with gold trim helmets. They cost 2 steel and 1 cloth each and the unique thing about these is that upon putting them on and right clicking the icon you can then select a decoration from the list provided, with each helmet having the same possible decorations, with over a dozen or so added, and which can be easily added to over time if anyone were to make their own crest designs.

## Why It's Good For The Game
I believe something like this would be appreciated by a lot of knight players with how it could better express their individuality or personality with a simple helmet decoration. Below is an example image.

## Pre-Merge Checklist


Stutternov:
Good PR, just fix conflicts and I'll set it to merge.

PR #617, opened 2024-08-29
Hair overhaul

NPC1314:
## About The Pull Request

* Adds a ton of hairs. Most are derivatives of TG hairs about 2/3 of the ones we have are already that though, some new, all curated. 3 brand new ones from Edharg
* The races all get a couple unique hairstyles for spice.
* Filepath, icon name, display name now all matches so you can actually find the damn thing you want
* Adds new +hidden extension for slimmer hairs and hoods since hidehair flag kinda broken. Doesnt do anything until added to specific hats/helmets for zero hair.
hairs

Any problems? 
-Will force everyone to select hair anew for all characters since savefiles will point to the old paths
-Some got renamed to fit the system, challenged players might think their hair has been deleted and cry
- A few of the elf specific hairs that got turned into generic got back to elves, but you got a dozen or more new otheres instead of those two for non elfs so no doubt its a big net gain.

Lot more hairs
More flavor to each race by letting them keep a few goodies each
Finally structure in the hair files

## Pre-Merge Checklist


bowlofcereal:
no more must all women look almost identical to eachother..

SilvernKell:
Hi! Shrinekeeper is missing from half-elf when it was available before! Please re-add it? Thanks

SilvernKell:
..Additionally, for women I can randomize a lot of hairs that are inaccessible on the list for the character's race. "Amazon" "Orc" being examples. Checking all races reveals that only aasimar can pick Amazon and martial, so it's listed, but "Orc" is not listed for any race. 
So you can randomize any hair you want anyway, but, I don't really understand setting some hairs to specific races. Martial for example would work just fine for a humen barbarian for example. And Junia, the tiefling specific hair, is almost identical to Lady and Wisp so what does locking it to tiefling provide, nobody would really notice, it doesn't really provide anything cultural.


An actual bug, Tiefling hair colour does not save correctly per character. It always defaults to an unselectable off-black colour instead of the 4 you can choose.

ThePainkiller:
Good for a fullmerge now that issues have been addressed and closed?

Stutternov (approved):

PR #627, opened 2024-08-31
Fixes smithed weapons losing their Wdefense modifier when Wielded and then Unwielded.

saltprjak:
## About The Pull Request

Smithed weapons used to return to their initial, unmodified wdefense value when unwielded, causing them to permanently lose their defensive bonuses. Now instead of returning it to the value, it detracts a flat 1 - same as how it adds a flat 1 when you wield an item.
Tested in a variety of ways - unwielded, dropped voluntarily, dropped out of stun, handed to other characters, stuck inside and ripped out, put into inventory and washing bins. Seems to work just fine.
gigaslop

## Why It's Good For The Game

Fixing Good

## Pre-Merge Checklist


code/game/objects/items.dm
```diff
@@ -1159,7 +1159,7 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 		wielded = FALSE
 		if(force_wielded)
 			force = initial(force)
-		wdefense = initial(wdefense)
+		wdefense = wdefense - 1
```
ThePainkiller, line 1162: Are you sure this shouldn't be -3 ? Weapons gain a +3 flat wdefense on wielding, this would just give 1 free wdefense

Cre77: this is the unwield proc! also wdefense from wielding got nerfed to 1.
again, i tested this in a variety of ways, the fix works.

ThePainkiller:
I need to have a word with whoever both did the PR that reduced wdefense when wielded to flat +1, and with who approved that :scary:

ThePainkiller (approved):
Whoever lowered it to 1 without ever revising the entire wdefense spread of weapons (like cutlass being one handed messers with high defense to compensate) will suffer the wrath of the hammer of Malum.

PR #628, opened 2024-08-31
Nets

SELFHELL:
## About The Pull Request

Adds throwing nets to the game,
Upon throwing them, they slow people down and reduce their SPD/END because there is a.. net on them

## Why It's Good For The Game

More Nonlethal options.

## Pre-Merge Checklist


NPC1314:
Seems like a pretty good idea to at least try. If it turns into cancer and everyone and their mother living and dying by the net its pretty contained and can be nerfed or nuked easy nuff. Id say a TM when the queue is cleared up a bit.

Biggest potential issue might actually be the craftability, do we want it to be so freely available? Maybe have it non craftable? Locked to some class? Require a trait to use?  Hmm.....

code/datums/status_effects/rogue/debuff.dm
```diff
@@ -295,6 +295,26 @@
...
 
+/atom/movable/screen/alert/status_effect/debuff/netted
+	name = "Net"
+	desc = "<span class='boldwarning'>A net was thrown on me.. how can I move?</span>\n"
+	icon_state = "muscles"
+
+/datum/status_effect/debuff/netted
+	id = "net"
+	alert_type = /atom/movable/screen/alert/status_effect/debuff/netted
+	effectedstats = list("speed" = -7, "endurance" = -5)
```
ThePainkiller, line 306:
```suggestion
	effectedstats = list("speed" = -3, "endurance" = -3)
```
Son these stat debuffs are INSANE. Tune them down a bit so we can do a bit of live testing with a TM, please.

code/modules/roguetown/roguecrafting/items.dm
```diff
@@ -59,6 +59,15 @@
 	verbage = "braid"
 	verbage_tp = "braids"
 
+/datum/crafting_recipe/roguetown/net
+	name = "net"
+	result = /obj/item/net
+	craftdiff = 2
+	reqs = list(/obj/item/rope = 2)
```
ThePainkiller, line 66: I would suggest anything else as a "weight" for the net to properly ensnare someone. Three rocks should do.

ThePainkiller (dismissed):

code/datums/status_effects/rogue/debuff.dm
SELFHELL: That was the goal, they are pretty easy to remove. But sure

ThePainkiller:
Remember it's better to adjust numbers up than down, with some testing on live we could gather some data on their use by players.

code/modules/roguetown/roguecrafting/items.dm
SELFHELL: Done

ThePainkiller:
Gonna be testmerging this one so that people can test their use, and after live feedback we can see whether the values need adjusting or any issues regarding buff removal due to tickrate arise.

ThePainkiller (approved):
I think they're safe to merge now with the slowdown + the slowdown trait helps establish a framework for a Town Guard idea Rebel had about mage garrison guard.

PR #635, opened 2024-09-02
Buffs the humen stat array

1N7Slayer:
## About The Pull Request
+1 to CON for both genders. This is a total of +3, equal to elves with +3. As they are the chosen race, plus it's just CON anyway, not a gamechanging stat like SPD, I think this is fair to bring their parity up to the other races, as they previously fell short with no meaningful upsides.

## Why It's Good For The Game

It's balance and lore represented fluff, what more could you ask for? #humensweep

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/roguetown/human/humen.dm
+++ b/code/modules/mob/living/carbon/human/species_types/roguetown/human/humen.dm
@@ -41,8 +41,8 @@
 	OFFSET_FACE_F = list(0,-1), OFFSET_BELT_F = list(0,0), OFFSET_BACK_F = list(0,-1), \
 	OFFSET_NECK_F = list(0,-1), OFFSET_MOUTH_F = list(0,-1), OFFSET_PANTS_F = list(0,0), \
 	OFFSET_SHIRT_F = list(0,0), OFFSET_ARMOR_F = list(0,0), OFFSET_UNDIES_F = list(0,0))
-	specstats = list("strength" = 0, "perception" = 0, "intelligence" = 0, "constitution" = 2, "endurance" = 2, "speed" = 0, "fortune" = 0)
-	specstats_f = list("strength" = 0, "perception" = 0, "intelligence" = 1, "constitution" = 1, "endurance" = 1, "speed" = 1, "fortune" = 0)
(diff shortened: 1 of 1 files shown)


Cre77:
Macha's rebalance police was to keep "equal exchange" for a race (with the quirk that str is considered as virtually two points), with a bonus +2 stat points added depending on gender. That's how it was with the old race rebalance, and that's how it was with the recent aasimar and tiefling rebalance. Take male dwarves for example - 1 STR (2 points), -2 PER, 2 CON, 2 END and - 2 SPD rounds them out nicely to a balance of +2 that they get from gender.
With humans already having the widest job availability spread, no downsides, and an established supremacy in-lore, I don't think they need free +2 stat points over every single other race.

ThePainkiller:
No nerfing endurance, I like my fucking limited stamina pool already.

Cre77:
> > Macha's rebalance policy was to keep "equal exchange" for a race (with the quirk that str is considered as virtually two points), with a bonus +2 stat points added depending on gender - CON and END for men, INT and SPD for women. That's how it was with the old race rebalance, and that's how it was with the recent aasimar and tiefling rebalance. Take male dwarves for example - 1 STR (2 points), -2 PER, 2 CON, 2 END and - 2 SPD rounds them out nicely to a balance of +2 that they get from gender. With humans already having the widest job availability spread, no downsides, and an established supremacy in-lore, I don't think they need free +2 stat points over every single other race.
> 
> They also have no specialization which makes them worse at everything, really, and don't get any unique things like their own language, dark vision, or free skills like reading for elves or mining for dwarves.

So your problem is that the specifically generalist race that's supposed to have no downsides is harder to minmax than non-generalist races? I don't think that's a problem to begin with.
As with darkvision, free skills and language - most of the time it's *pure flavor* that never comes up compared to the stats. If you want, you can easily look into propping them up with languages and gimmicks, but this is just a reactionary solution to an imaginary problem.

DovaKitty (changes requested):
The current concept for buffing Humen is to give them a choice of +1 to whatever stat they want and +1 fortune. Fortune does very little in the current code, so it does not compare to the +1 con proposed here. If you can find a way to implement the choice of stat on the character setup page, that would be great, but if you think you can't I'd recommend closing this PR until I can get around to doing it myself.

Guayo has given down the order that all races are to get certain unique buffs, and while this would be a buff, it would hardly be 'unique'.

1N7Slayer:
sounds good for now then

darn

PR #642, opened 2024-09-02
New Armor And Clothing

WWWWW-Git:
Adds new armor and clothing. The majority of it can be crafted via sewing, weaving and smithing, but some of it is not currently obtainable via normal means. Below is images of the proposed armors.

## Why It's Good For The Game

Someone once said that variety was the spice of life. This is true, and I believe the addition of these new armors and clothes will truly be able to further spice up roleplaying possibilities within the server.


sanshoom:
Can you name which pieces you wouldn't normally be able to get without a badmin spawning it in for you?

sanshoom:
> > Can you name which pieces you wouldn't normally be able to get without a badmin spawning it in for you?
> 
> Valorian Blood Baro set and the Graggar helmet.

You /could/ put the Graggar helmet into the Graggarite Assassins satchel/backpack, but I would probably ask Dova if he would even want that beforehand, since he made the assassins.

PR #667, opened 2024-09-05
You now perma-die slower when in bloodloss crit

saltprjak:
## About The Pull Request

Putting this up here for visibility in discussions, since it got largely buried in contribs chat.

This change simply makes it so that after passing out and falling over from bloodloss you won't die as fast. This goes from ~11 oxyloss to 4, meaning you now die almost three times slower

## Why It's Good For The Game

A torn artery usually means death unless there's a miracle user nearby - regardless of sewing and waterboarding, and especially if it's fresh off a fight. This hopefully should make people permadie less often, and promote easier captures to extend people's roleplay, rescue injured more readily, be able to leave bled-out opponents to be found or saved by others, etc.. This does NOT make bloodloss any less lethal, it just makes it so that you don't permadie faster than you unlock the ability to willingly succumb.

To all speeches of realism (aside from the fact that it takes longer to die from such injuries to begin with - or that in some other cases you would die instantly and irreversibly) - I raise **SUSPENSION OF DISBELIEF**. This is done purely to let people stick around longer, especially so when other players WANT them to. Anyone else can still hit the death button after two minutes to dramatically die or shove off to respawn and metagame. (i'm also pretty sure people only ever started dying after passing out so fast because of the medical pr).

You still CANNOT outheal extreme ongoing bloodloss by sleeping or drinking water, and you CANNOT keep yourself alive longer by "sleeping" while already Dying. This DOES NOT affect anything else like drowning or poisoning.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/weapons.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/weapons.dm
@@ -114,7 +114,7 @@
 	appro_skill = /datum/skill/craft/engineering
 	req_bar = /obj/item/ingot/iron
 	additional_items = list(/obj/item/grown/log/tree/stick)
-	created_item = /obj/projectile/bullet/reusable/arrow
+	created_item = /obj/item/ammo_casing/caseless/rogue/arrow
 	createmultiple = TRUE
 	createditem_num = 4
 	i_type = "Ammo"


ThePainkiller:
Do remember to account for ticklag, as the value difference is way too high. Adjust for a middle ground between current and this PR, and we'll testmerge to see how it plays in practice.
I'm fearing people being put on permacrit for half an hour with no chance to properly die.

Cre77:
> Do remember to account for ticklag, as the value difference is way too high. Adjust for a middle ground between current and this PR, and we'll testmerge to see how it plays in practice. I'm fearing people being put on permacrit for half an hour with no chance to properly die.

There is a button that unlocks after two minutes floating right in the middle of the screen - you just die before it ever goes off required duration.

Cre77:
upped to 4 per tick on request

1N7Slayer:
ding ding ding ding ding based PR

Stutternov (approved):

PR #663, opened 2024-09-05
Fixes an issue with Pestran acolytes not spawning with any actual miracles or outfit. Fixes an issue with clerics spawning with armor they are not proficient with, gives them chainmail shirt instead for identical coverage.

1N7Slayer:
## About The Pull Request
Merge priority: urgent bugfix. Pestran acolytes spawn with less clothing than beggars and no miracles to show for it.

## Why It's Good For The Game

Bugfix.

## Pre-Merge Checklist

- [x] You tested this on a local server.
- [x] This code did not runtime during testing.
- [x] You documented all of your changes.

PR #674, opened 2024-09-06
Restores Cleric's Half-Plate

DovaKitty:
## About The Pull Request
* Replaces Cleric's chainmail armor back to half-plate.

## Why It's Good For The Game
#663 was a bugfix PR that swapped Cleric's halfplate for a chainshirt under the erroneous belief that half-plate was heavy and not medium armor. The PR was up for less than an hour, so this was not caught and players could not give their opinions. Balance wise, the chain shirt and half-plate are fairly close, but I think it's worthwhile to discuss the change in a new PR to decide the issue. Feel free to post your thoughts in the comments. 

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/cleric.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/cleric.dm
@@ -47,7 +47,7 @@
 			wrists = /obj/item/clothing/neck/roguetown/psicross/pestra
 			cloak = /obj/item/clothing/cloak/stabard/templar/pestra
 
-	armor = /obj/item/clothing/suit/roguetown/armor/chainmail
+	armor = /obj/item/clothing/suit/roguetown/armor/plate
 	shirt = /obj/item/clothing/suit/roguetown/shirt/shortshirt/random
 	pants = /obj/item/clothing/under/roguetown/trou/leather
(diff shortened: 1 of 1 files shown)


YuiY1997:
I would rather keep the halfplate.

1. I think cleric in its current state has been over corrected
2. If balance is similar, why change it? 
3. I think clerics having half-plate is cool and adds a bit of flavor. I really do not think cleric is at all overpowered or overgeared at this point. We should have just stuck with the halfplate.

Optimism333:
They were fine as they were, imo

Optimism333 (approved):
Would have preferred more input before this was changed. Clerics having halfplate is fine, I suppose. I was thinking about giving them an iron breastplate over chain, but halfplate alone is also fine.

ThePainkiller:
A reminder that "input" on cases of balance is a suggestion at most, and maintainers handle balance above everyone else, you have it spelled out clearly on the PR template whenever a pull request is pushed :fortune:

Stutternov (approved):
While I may have an internalized hatred for Clerics, this is getting merged as that chainmail change seems accidental or was at least not properly labeled; and was speed merged.

I don't blame anyone, just saying if we want to remove something like that balance wise it'll need to be labeled and discussed - this merging this because it restores it.

Good catch.

PR #675, opened 2024-09-06
Churn undead tweak: Necra enjoyer edition.

YuiY1997:
## About The Pull Request

Doubles the chance for churn undead to work (It's not that crazy I promise please read)

## Why It's Good For The Game

So, there was a previous rebalance of turn undead that made it so players no longer had a negative chance for turn undead to actually work against enemies.

Unfortunately, churn undead is still next to useless for clerics of Necra, who are supposed to be specialized in combating the undead. Presently it has only a 40% chance of it working on regular undead (skeletons, zombies), and a 20% chance of it working on vampires.

Since each level previously added a 40% chance to do damage, all classes with level 3 faith or above (acolytes and other church roles) already had a 120% chance to do damage with churn undead, and 100% after the 20% reduction against vampires. Either way it is guaranteed that they do damage.

Now, clerics of necra (who are the only clerics who get churn undead, and have no other combat spells) have a base 80% chance, which goes down to 60% against vampires.

This still keeps clerics in a place worse than the 100% chance provided by acolytes and actual church roles (that they had before and after this buff to the spell) while still allowing them to be undead-fighting specialists.

It is also worth noting that Churn Undead only works against un-disguised lesser vampires, and Necra clerics have no way to force the loss of a disguise since they have no cleansing flame. This makes the undead-fighting specialists objectively worse at fighting the undead than astratan clerics without this change.

I believe this change effectively makes clerics of necra useful in the one field they are supposed to be while also keeping them magically inferior to church roles.

## Pre-Merge Checklist


ThePainkiller:
Some good points to be made, we might have to make a spell and assign it to clerics to rebuke a disguise from a vampire somehow, without the whole "lighting them on fire" thing the astratan version does.
It would be a good idea to make Necran followers have actual spells that deal against the undead, to give them their own niche.
I'll review the chances in a bit.

YuiY1997:
>It would be a good idea to make Necran followers have actual spells that deal against the undead, to give them their own niche. I'll review the chances in a bit.

This would actually be a great thing to give as a t3 spell for acolytes of necra, since they have none after speak with dead was removed.

ThePainkiller (approved):
Pending a review of Pestran spells to make them better suited at hunting vampires and dealing with them, without cucking Astratans too much.

PR #485, opened 2024-08-07
Dark Vision for Dark Elf and Lesser for Half-Elf and Elf

BadAtThisGame302:
## About The Pull Request

Ports this PR, https://github.com/Blackstone-SS13/BLACKSTONE/pull/135, adding Dark Vision to all Elvish races, lesser so for halflings and snow elves.

## Why It's Good For The Game

More racially unique traits, here are pics down below of what it looks like.

Half Elf and Elf
image

Dark Elf
image

## Pre-Merge Checklist

- [x] You tested this on a local server.
- [x] This code did not runtime during testing.
- [x] You documented all of your changes.

PR #680, opened 2024-09-08
Aasimar Unique Traits (Much Higher Nutrition Needs, Slightly Less Pain and Bleeding)

Optimism333:
## About The Pull Request
Following the spirit of #485 in giving races more unique mechanics, I decided to give aasimar a significantly higher hunger rate than the other races, but slightly less pain and bleeding from injuries to reflect their construction as a soldier race. I did my best to make sure that these changes were reasonably balanced but also in line with the flavor of the race.

I also adjusted female aasimar stats slightly. They now have a -2 in Intelligence instead of -1, but a +1 to Endurance instead of 0. The only difference now between male and female aasimar now is that females swap a point in Constitution for one in Speed.

### Changes Summarized ###
- Aasimar have 200% nutrition loss, compared to the other races. (Will see how this plays in-game. In my testing hunger still seemed more than manageable, but it might need adjusting)
- Aasimar experience 10% less pain from wounds
- Aasimar bleed 20% less from wounds
- Female aasimar now have -2 to Intelligence instead of -1, but +1 in Endurance instead of 0.

### Did You Test It ###
- Tested every different variable added in this pr (nutrition_mod, bleed_mod, pain_mod), and they all appear to function as intended. 

## Why It's Good For The Game
- More unique racial flavor is good, provided that an effort is made to respect balance. I think that aasimar needing to eat much more than other races is fitting, and the slight reduction to pain and bleeding gives them a small edge in combat, but a serious injury will still put them down easily enough.

## Pre-Merge Checklist


ThePainkiller:
Oh yes, these seem perfect traits for them

code/modules/mob/living/carbon/human/species.dm
```diff
@@ -64,6 +65,8 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 	var/coldmod = 1		// multiplier for cold damage
 	var/heatmod = 1		// multiplier for heat damage
 	var/stunmod = 1		// multiplier for stun duration
+	var/bleed_mod = 1	// multiplier for blood loss
```
ThePainkiller, line 68: This var right here sounds wonderful for things that should slow down bleeding but not staunch it, e.g tourniquets or bandages...interesting

ThePainkiller (approved):
Overall solid spread of drawback + boon, and quite fitting for indefatigable immortal soldiers

PR #681, opened 2024-09-08
Sellable organs, with variable price based on stats

saltprjak:
## About The Pull Request

Organs now get a base price of 10, limbs get 5. This price changes along with stat modification, with things only really getting expensive at over 13~. Additionally removes feldsher's mask RNG - they are now guaranteed to get it and the hood. Tongue is no longer invisible.

Out of issues during testing, mobs that do not have their stats modified at all and do not have time to have them modified via moodlets or debuffs (i.e. - gobs, orcs) will have their organs and limbs at the "default" price. Tackling this would prob require an extensive rewrite which I can't be bothered with. Not much of an issue, considering organ trade is oriented at the much more varied player characters.

## Why It's Good For The Game

Lets people drag bodies off to the Feldsher to have them butchered for sellable organs. Mundane ones aren't too profitable, but those with high stats might make it worth it to actively stage accidents and place hits on people with expensive innards.
Organ price is affected by ALL stat changes, which can both be of help and misfortune. Bled-out bodies will be worth less (albeit you can remedy that with the blood parasite - i tested it), but at the same time you can feed still-living people good food, drugs and potions to prepare them for dissection. Stats affect things as such:
Brain - INT
Eyes - PER
Tongue - LUC
Arms - STR
Heart, Lungs - END
Liver, Stomach, Guts - CON
Legs - SPD

Combined "cheap" stuff will net you around a 100~ (if at the cost of a lot of time), maxed out stat individual organs can top out at just under 200 - but, again, the latter requires someone important to die, or for the feldsher to find someone living and do some really creative work with stat-stacking.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/assassin.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/rare/assassin.dm
@@ -134,29 +134,39 @@
 			backl = /obj/item/storage/backpack/rogue/backpack
 			backpack_contents = list(/obj/item/flint = 1, /obj/item/rogueweapon/huntingknife = 1)
 		if("Noble")
+			var/prev_real_name = H.real_name
+			var/prev_name = H.name
+			var/honorary = "Lord"
+			if(H.gender == FEMALE)
+				honorary = "Lady"
+			H.real_name = "[honorary] [prev_real_name]"
+			H.name = "[honorary] [prev_name]"
+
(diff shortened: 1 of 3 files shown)


1N7Slayer:
another banger cre7 pr

Cre77:
i'm also considering giving feldsher steelhearted since they freak out over the very limbs they amputate. thoughts on that?

ThePainkiller (approved):
This was supposed to be merged long ago, I apologise.

PR #683, opened 2024-09-09
Fixes Decorated Helmets having the wrong FOV blocker

saltprjak:
## About The Pull Request

Decorated helmets now give you harsher tunnel vision like all other closed ones.

## Why It's Good For The Game

Drip does not open your third eye, surprisingly. Kings seem to have taken to rushing decorated helmets for themselves and everyone they like since they have no downsides.

## Pre-Merge Checklist


beepbopbeepbop:
It would also be good to fix the fact their visors can't be flipped up, to account for this.

Cre77:
> It would also be good to fix the fact their visors can't be flipped up, to account for this.

that would be a whole different thing to rebalance - and also fix. you're not really getting as much value-per-ingot out of heavy helmets as you would with a chain coif (or simple helmet) and a steel mask, the FOV block lets people bypass your parries when attacking from the side (the FOV_behind, curiously, does not have this limitation, those diagonal from you can still be parried), and flipping visors is REALLY buggy.

this just knocks down the objectively better helmet option down a notch, and aligns it with other closed helmets.

ThePainkiller (approved):
Thank you, I was going to tackle this one but it slipped my mind.

The whole visor up and down thing will have to be sprited by whoever created these helmets though, as that requires 2 additional sprites (icon and on-mob), not just code adjustment

For now they're all bucket helmets with no toggleable visors, and should get their FOV as every other helmet of their family does.

PR #706, opened 2024-09-11
Ports Ratwood's Feint Rework, lessens Bandit INT malus

saltprjak:
## About The Pull Request

Ports https://github.com/Rotwood-Vale/Ratwood-Keep/pull/756 by ddsaw, with added status alerts for its respective effects.
Also reduces the pretty pointless intelligence malus from bandits, to prevent feint spam being the go-to way to shut them down.

Original PR's changes are as follows, copypasted:
Reworks feint to be based on int difference and to a lesser degree spd difference. Changes the impact of feint intent to instead grant a longer immobilize effect (still barely notable but very hard to test on a local server, its still too quick to tab between two clients and see if I can move)
The actual number changes:

SKILL IMPACT 10% or 15% if feint intent -> always 15%
INT IMPACT Singular check to see if the opponent has more int -> int difference *10
SPD IMPACT None -> 5% per difference
Feint Intent -> Increases Impact of Skill difference -> Increases duration of Immobilize
## Why It's Good For The Game

Gives INT races and characters an angle of viability in melee combat, and offsets intelligence-weak STR characters' power with a weakness to this select intent. Gives a new option for people ganging up on powerful characters that isn't just grab spam.

Never really understood why the bandit intelligence malus was kept around to begin with, aside from flavor.

## Pre-Merge Checklist

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -47,6 +47,7 @@
 #define TRAIT_VAMPMANSION "vampiremansion"
 #define TRAIT_LIMPDICK "limp_dick"
 #define TRAIT_SEXPASS "sexpass"
+#define TRAIT_INTRAINING "intraining" //allows certain roles to bypass the average skill limitation of training dummies
 #define TRAIT_STEELHEARTED "steelhearted" //no bad mood from dismembering or seeing this
 #define TRAIT_IWASREVIVED "iwasrevived" //prevents PQ gain from reviving the same person twice
 #define TRAIT_IWASUNZOMBIFIED "iwasunzombified" //prevents PQ gain from curing a zombie twice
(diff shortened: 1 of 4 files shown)


ThePainkiller:
While I agree that bandits' INT malus was too high, I'd still rather keep it at a -1 for their whole savagery and connection to Matthios
Given that they now have a system to actually summon equipment, it's not like having a debuff for crafting affects them that much.

code/game/objects/items/rogueweapons/rmb_intents.dm
```diff
@@ -80,11 +83,22 @@
 
 /datum/status_effect/debuff/feinted
 	id = "nofeint"
-	duration = 50
+	alert_type = /atom/movable/screen/alert/status_effect/debuff/feinted
+	duration = 5 SECONDS
```
ThePainkiller, line 87: On testing alchemy, using the duration in exact deciseconds has proven more reliable in making buffs/debuffs fade away than using pre-built definitions like `X SECONDS/MINUTES`, so I suggest reverting these just to err on the side of caution.

code/modules/antagonists/roguetown/villain/bandit.dm
DovaKitty, line 1: Feint code looks good. This should be reduced to -1 instead of removed altogether. Bandits know how to read for some reason, but I'd not say they're as intelligence as the average towner.

DovaKitty (changes requested):
Feint is good, moving Bandit to -1 would be okay.

code/game/objects/items/rogueweapons/rmb_intents.dm
Cre77: changes addressed

NPC1314:
Don't see any reason this shouldn't get a test since comments adressed.

NPC1314 (approved):

PR #708, opened 2024-09-12
Fixes Longbeard Hammer (Balance, Bugfix)

Optimism333:
## About The Pull Request
Longbeard hammer was essentially broken, and its damage values were PITIFUL. It's been made a subtype of the steel grand mace, so it now actually hits respectably hard, and I cleaned up the code a bit.

## Why It's Good For The Game
A two-handed dwarven warhammer shouldn't hit like a wooden club, and less code clutter is good.

## Pre-Merge Checklist

### DIFF
--- a/code/game/objects/items/rogueitems/waterskins.dm
+++ b/code/game/objects/items/rogueitems/waterskins.dm
@@ -16,3 +16,6 @@
 	fillsounds = list('sound/items/fillcup.ogg')
 	poursounds = list('sound/items/fillbottle.ogg')
 	sewrepair = TRUE
+
+/obj/item/reagent_containers/glass/bottle/waterskin/milk // Filled subtype used by the cheesemaker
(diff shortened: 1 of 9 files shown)


Optimism333:
> Could be stronger. I tested it and it still hits like a wiffle ball bat. Otherwise, good PR!

As a child of the grand mace, it's got 35 force when wielded in both hands. Personally I think that's more than alright, but I'm open to more feedback on that.

ThePainkiller:
Making it a godendag child does feel like a nerf, I believe it should be more on par with the Eagle's Beak to reach its proper level

Optimism333:
> Making it a godendag child does feel like a nerf, I believe it should be more on par with the Eagle's Beak to reach its proper level

It's not a nerf at all. It had 20 force previously when wielded, the same as a copper mace. As it stands it has 35 now when wielded. But if you want a higher value, I can do that.

ThePainkiller (approved):

PR #709, opened 2024-09-12
Blunt attack auto-knockdowns no longer drop your weapon, or hard-stun you.

saltprjak:
## About The Pull Request

Title. The Butterfinger PR made it so that getting knocked down no longer made you drop your weapon - this moves blunt attack knockdowns from Paralyze to actual Knockdown.
Also halves the duration of that effect - which only serves to let you get up sooner, you can still act while under the knockdown effect.

## Why It's Good For The Game

Previously, if you had enough damage, you could make someone drop their weapon while also keeping them hard-stunned for as long as you can keep attacking. Ideally, the whole knockdown-on-attack could use a rewrite, but this change should move blunt combat away from stun prevalence and make fighting mace-wielding swift-intent VLs less obnoxious.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -2866,4 +2866,4 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 		skill_modifier = mind.get_skill_level(/datum/skill/misc/athletics)
 	var/modifier = -distance
 	if(!prob(STASPD+skill_modifier+modifier))
-		Paralyze(15)
+		Knockdown(8)


1N7Slayer:
LGTM, was pouring over trying to look for this earlier and didn't think to check for on-hit paralysis sources.

This is a pretty critical balance issue that makes maces way stronger than they need to be -- hopefully by removing their disarm on hit maces can get buffed back in their other statistics.

NPC1314:
We'll see how it works out

ThePainkiller:
I don't really like the whole "you can act while stunned" part to be honest, that's going to make taking anyone alive and chain them even more of a pain in the ass if they can still crawl.

NPC1314 (approved):

PR #710, opened 2024-09-12
Lesser Heal specialization and conditional buffs

BadAtThisGame302:
## About The Pull Request

Ports this PR, https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/254, which specialized lesser miracle for each individual God/Saint/Force even ~~Psydon~~ and Zizo and the rest of the evil gang and gives them conditional buffs depending of each Gods affinity.

## Why It's Good For The Game

Huge QOL change for the Church and I would love to see how this plays out

## Pre-Merge Checklist


ThePainkiller:
While this is good overall and adds a ton of flavor, I would prefer if WOUNDS being healed be moved completely to a Pestra-oriented heal, to make Physikers more sought after, and so that a heal button is not an actual semi-aheal for people so easily.

NPC1314:
This has been discussed and consensus is to limit woundheal for minor healing to Pestra and Astrata while we handle lowpop issues with splints so theres a way to handle broken ribs and arms. 

Do that adjustment (wound only Astrata and Pestra) and this is good for a testmerge.

NPC1314:
Right this has been on hold long enought, its just a straight upgrade of the current spell, the planned splint and aid spells arent going to materialize any time soon so might as well play with the better version with all the cool bling until that happens.

NPC1314 (approved):

PR #713, opened 2024-09-13
Zybantu Forever!

CormanQuewye:
## About The Pull Request

- Returns the Zybantine merc back to their excellent, original drip instead of the bulbous, plate knock-off they currently have;
- Returns their trousers to being black instead of the eye-burning, bright red they currently have (but were given a purple shirt so they're not totally colourless);
- Gives them Shield 2 instead of a random selection and takes away their Bow skill (why did they have bows to begin with?)

## Why It's Good For The Game

The current armour sprite, especially with the bright red pants, are actually difficult to look at and didn't really have much reason to swap to. The previous armour fit the kit, it looked good, this was a needless change that took away the main reason to pick Zybantine, which is the fashinable set. Also, the new set they were given for some reason lacked leg armour (despite the sprite clearly having a knee-length skirt?) which gave the merc class focused on melee fighting the glaring weakness of everything below the waist, considering they do not spawn with armoured leggings.

I also removed the shield skill randomness, setting it to 2, but took away their bow skill, since I saw little reason for it being there in the first place.

For visual comparison, here's the original Zybantine look (which this PR reverts to):
image_2024-09-13_055235367

Here's what Zybantine's current set looks like:
image_2024-09-13_055404241

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/other/merc_classes/zybantine.dm
+++ b/code/modules/jobs/job_types/roguetown/other/merc_classes/zybantine.dm
@@ -23,11 +23,11 @@
 	head = /obj/item/clothing/head/roguetown/helmet/sallet/zybantine
 	gloves = /obj/item/clothing/gloves/roguetown/angle
 	belt = /obj/item/storage/belt/rogue/leather/shalal
-	armor = /obj/item/clothing/suit/roguetown/armor/brigandine/coatplates
+	armor = /obj/item/clothing/suit/roguetown/armor/plate/scale
 	beltr = /obj/item/rogueweapon/sword/long/rider
 	beltl= /obj/item/flashlight/flare/torch/lantern
-	shirt = /obj/item/clothing/suit/roguetown/shirt/undershirt/black
-	pants = /obj/item/clothing/under/roguetown/tights/red
+	shirt = /obj/item/clothing/suit/roguetown/shirt/undershirt/purple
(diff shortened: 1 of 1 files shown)


ThePainkiller:
Already offered the alternative to add a different sprite to this armor, which isn't just pasting the Scalemail sprite on it, that covers the legs, with the exact same values and protections unchanged, so that they can have both their desired drip and their functional leg protection, but I will not allow non-limited classes to spawn with Scalemail-equivalent equipment. The Veteran spawns with one, and Zybantu mercs should not.

CormanQuewye:
> Already offered the alternative to add a different sprite to this armor, which isn't just pasting the Scalemail sprite on it, that covers the legs, with the exact same values and protections unchanged, so that they can have both their desired drip and their functional leg protection, but I will not allow non-limited classes to spawn with Scalemail-equivalent equipment. The Veteran spawns with one, and Zybantu mercs should not.

As said, I offered an alternative that would not require putting in ten times more effort than you did on the Zybantine change where we can swap the numerical values of Coat of Plates and Scalemail. This way there wouldn't be the bloat you claim to want to avoid, the sprite (a purely visual item) would be associated with a weaker armour, you can treat Coat of Plates in the way you were going to treat Scalemail.

The idea that I should have to sprite something you did not and monopolized an already used sprite for your own ends is absolutely bonkers.

PR #719, opened 2024-09-14
Fixes Bandit examine "job" and double-credit

saltprjak:
## About The Pull Request

Fixes Bandits showing up without a "job" next to their race on examining, they're drifters now.
Bandits also no longer show up twice in the credits section.

image

They still have their old pre-selection name during the credits sequence (but not in the greentext results logs), if anyone knows how that might be fixed - please!!

## Why It's Good For The Game

humbug

## Pre-Merge Checklist


BadAtThisGame302 (approved):
Thanks alot for this fix.

ThePainkiller (approved):
Let's give a full merge for now to stop people metaing and we'll see about fixing the later issue down the line. It might have something to do with using `real_name` rather than `name`.

PR #720, opened 2024-09-14
Allow Aasimar to be Warriors

BastardizedCoffee:
I have absolutely no idea why this was that way. I won't even bother explaining.

## About The Pull Request

Changes 1 line in the Warrior class file to let Aasimar be warriors.

## Why It's Good For The Game

I have no idea why this wasn't a thing. This is stupid and this is a salt PR.

## Pre-Merge Checklist


ThePainkiller:
I don't want to quote opti, but I believe they were barred from some combat classes out of how good their stats (and now inherent passives) were, so I'd like to consult him first on that regard of balance before making a decision.

Optimism333 (approved):
Warrior isn't a rare combat class and doesn't have crazy stat bonuses. I think this is alright, particularly since they can be clerics as is.

ThePainkiller (approved):
Approved by Opti, no qualms about it

PR #734, opened 2024-09-16
Deadite

NPC1314:
## About The Pull Request

Basically a nerf for deadites, or a fix, take your pick.
Their emoting was buggy so made a new to make sure their deadite moans actually play, though less spammy now
Added stinky cloud to them to help identify them, mask or no mask
Actually get shittier stats (slow and cant sprint so much, kinda dumb) previously was unchanged by death
Duplicate Idle emote deleted, didnt do anything 

## Pre-Merge Checklist


ThePainkiller:
BASED BASED BASED BASED BASED BASED BASED BASED BASED BASED BASED BASED

Optimism333 (commented):
Personally I think the speed nerf is probably sufficient. As long as they can be outrun, they will already be a significantly lesser threat.

I don't think they need even lower constitution on top of that.

NPC1314:
> Personally I think the speed nerf is probably sufficient. As long as they can be outrun, they will already be a significantly lesser threat.
> 
> I don't think they need even lower constitution on top of that.

Sounds good, its not like theyre some unstoppable menace, just want them to be more zombie like and less just as normal with slightly greener skin. Con malus to prevent sprinting too much bascially, point isnt to cuckl them too badly. Ill yeet the con malus when I can find the branch its made one, things are messy

Stutternov (approved):

PR #739, opened 2024-09-17
Belting_Merc_Mains_actually_satchel4

NPC1314:
## About The Pull Request

_* Introduces the filter for submission holes/stockpiles. Mercs can't use them. Why? Mercs should not spend their rounds grinding resources and feeding it to a NPC vendor, their whole point is to sell their services to actual players and interact.
Lots of roles really shouldnt spend their rounds logging and stuffing the stockpile, but baby steps._
(commented out this part, might need refinement or rolled out as part of bigger package but not worth the time to debate it this week)

* Mercs get their key and complimentary bandage tucked into their belt now instead of having the key in their backpack.
The standard is to spawn keys on hip or belt, no reason to not keep that consistent.

* Satchels gets 4 slots instead of 3. Was discussed ages ago, again and again, but never happened, so doing it now.
Satchels should be bit more roomy than belts and 4 items isnt enough to bust inventory managment, should be good.l

* Grenzelhoft had doubled starting money, up to 60! on spawn, making them one of the richest in town on spawn. Fixing that oversight, and giving exact same starting cash as other mercs. 

EDIIT
Renamed to be slightly less confusing, the submission hole stuff is on hold  since the incoming economy stuff is a far better starting point

## Pre-Merge Checklist


1N7Slayer:
You may want to add this to garrison guards then.

lavillastrangiato:
What kind of problem is submission filter intended to fix? What if you just want to get some salt from the place you technically count as a resident of, and can't because it's a "cursed local contraption"?

ThePainkiller:
> What kind of problem is submission filter intended to fix? What if you just want to get some salt from the place you technically count as a resident of, and can't because it's a "cursed local contraption"?

You're mistaking the submission hole with the stockload, this is so that they cannot SELL anything and get money directly deposited on their bank accounts through selling items, this doesn't bar them from PURCHASING, say, salt, from the stockload AFAIK

NPC1314:
Actually it will prevent purchases as well, its going to be a lot harder to target just selling but not purchases, probably.

lavillastrangiato:
> You're mistaking the submission hole with the stockload, this is so that they cannot SELL anything and get money directly deposited on their bank accounts through selling items, this doesn't bar them from PURCHASING, say, salt, from the stockload AFAIK

No, I already read the code.

> Actually it will prevent purchases as well, its going to be a lot harder to target just selling but not purchases, probably.

Then this feels kind of like a PITA. I don't think you should "encourage" interaction by cutting someone off from easy supplies if they just want to make a bucket or salumoi.

crunchypretzels:
as a merc player; stock loading isn't even that much money. I do it occasionally to make enough to cover a crossbow license(if the king's doing that) or food, as we usually spawn with about 7 mammons unless you immediately rush to the Captain and beg for a job. If I can afford to live a little before I get a client, I'm less likely to always rush to the Keep.

ThePainkiller:
Adding it to the stockpile as well is a little bleak, this won't be well received by mercs
I understand the sentiment and how this'd goad them into actually getting coin by contracts but it's gonna make a lot of people butthurt.

NPC1314:
Making a more complicated filter is probably possible but not right now, only alternative is commenting out the filter for now and leave the problems for later, might as well.

NPC1314:
Ok its been like 2 weeks this is getting tested.

ThePainkiller (approved):
Considering no changes and the code is commented, on your request I'll have to merge this one fully then.

PR #751, opened 2024-09-18
Autumn Cleaning pt 1

NPC1314:
## About The Pull Request

Purges a chunk of commented out file left over from old cooking and hydroponics
Also cleans out the dupe apple/potato/onion stuff in hydro, the TG seed vendor and some TG cargo bounty stuff associated with this. Also couple more depreceated hydro veggie files and some unused TG crafting recipes (for robots etc)

Has no effect on gameplay. Reduces duplicates and trash files. Faster & less clutter when looking in VSC.
Tested, no issues.

## Pre-Merge Checklist


code/datums/components/crafting/recipes.dm
```diff
@@ -24,7 +24,7 @@
 	var/craftdiff = 1
 	var/sellprice = 0
 	//crafting diff, every diff removes 25% chance to craft
-
+/*
```
ThePainkiller, line 27: Erase rather than comment down, please.

code/modules/food_and_drinks/recipes/processor_recipes.dm
```diff
@@ -3,7 +3,7 @@
 	var/output
 	var/time = 40
 	var/required_machine = /obj/machinery/processor
-
+/*
```
ThePainkiller, line 6: Ditto as above. Erase rather than comment out. SPECIALLY for line 9 which could get us TOS'd.

code/datums/components/crafting/recipes.dm
NPC1314: Figured it was worth keeping one of the TG style recipes just so people can see how the cat/subcat system worked, in case someone wanna do something with the RT recipes in trhe future

DorfGuy1:
are you going to get the old tg changlelog files from like 2014 too? Those throw up a tone of hits when code diving.

code/datums/components/crafting/recipes.dm
ThePainkiller: You could always just leave a commented out "example" recipe with code comments on all slots as deprecated doccumentation and just erase the rest of the unused block, less pointless data to keep around when only one good example suffices :P

NPC1314:
> are you going to get the old tg changlelog files from like 2014 too? Those throw up a tone of hits when code diving.

I might do in part 2 if I remember and get the time, but focus is cleaning up the systems and file structure I've been working most in, so not in this PR.

code/datums/components/crafting/recipes.dm
NPC1314: that exactly what it is though, I think youre mistaking this for the processor recipe page, thats already sorted. This one just has the example one, just changed the name to /testing now but its honestly as tight it can get, lets roll this bad boy out
firingpin

PR #753, opened 2024-09-18
[FIX (Sorta...)] Removes Veteran's Glitched Force-Flaw

SaintBurgher:
## About The Pull Request
Removes forced cyclops from Vet.

## Why It's Good For The Game
Currently forcing this flaw is glitched and just doesn't work properly. Plus people can always choose cyclops if they want.

Right now, Veteran is not a popular role and removing this opens more avenues for RP through taking other flaws. Pigeonholing cyclops is a bit of a stain on the role.

## Pre-Merge Checklist


sanshoom:
lame.

ThePainkiller (approved):
Code wise it was poorly handled yeah, and this opens up more character customization rather than cajoling the role into "that one-eyed old vet". People can still choose that flaw if they so wish, it's better not to just outright force it.

PR #757, opened 2024-09-18
Murderwoods

NPC1314:
## About The Pull Request

At some point maneaters and mancatchers got yeeted from spawnpool in forest. Reenables this.

Mapgen edit so cant be localtested.

Also two latejoin adv spawn spots added to bog

## Pre-Merge Checklist


AceVX:
Make murderwoods dangerous again oooraaah

Optimism333 (approved):
Bring murder back to the murderwoods...

ThePainkiller (approved):
Let's pray this makes the forests more dangerous like they used to be before open source

NPC1314:
If it turns out to be poo we know which mapgen to edit now at least, could tweak it all sorts of ways I guess but we gotta start somewhere

Zoeycore:
Could be fine with much lower spawnrates, but rn it feels quite bad.

The Murderwoods already has no reason to go there, outside of RP or spawning as Drifter. This change only amplifies that, and just makes RPing/PvPing in this zone pretty unbearable.

ThePainkiller:
The murderwoods were never a good place to RP, historically, it's always been a place that you try to exit as fast as you can to reach town or you'll get mugged by bandits or kidnapped by vampires.
Do they spawn TOO often? If so we can adjust the numbers on a later PR but unfortunately this is something we cannot test on localtesting and only ever works on live.

PR #774, opened 2024-09-22
Addresses an oversight in which monk had too much combat skill post-drifter combat skill changes.

1N7Slayer:
## About The Pull Request

Technically a balance PR but monk has infinite slots for the most part, and monk is most prone to the issue of "dying in the forest only to be a literally unbeatable zombie that can't be parried due to combat skill difference". Certain skills had their randomness removed, just to give them guaranteed non-combat usefulness. Polearms up to 3, makes them equally capable with staves, traditional polearms and their fists. Some stats were also dumbed down a bit to have parity with other drifter roles, while giving them just enough associated with a warrior-monk of Ravox to perform decently - brought their speed slightly up to make dodge expert more useful on them, as they cannot wear medium or better armor.

## Why It's Good For The Game

As mentioned - design philosophy parity. It's not a rare drifter role, it's a common one, and as such it should not exceed tier 3 combat skill unless it is a rare class or has significant drawbacks in other areas. If maints want this to be a rare instead I can revert certain changes and make it so. 

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/monk.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/monk.dm
@@ -32,19 +32,19 @@
 
 	if(H.mind)
 		H.mind.adjust_skillrank(/datum/skill/misc/reading, 3, TRUE)
-		H.mind.adjust_skillrank(/datum/skill/combat/unarmed, 5, TRUE)
-		H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 4, TRUE)
-		H.mind.adjust_skillrank(/datum/skill/combat/polearms, pick(1,1,2), TRUE) // Wood staff
+		H.mind.adjust_skillrank(/datum/skill/combat/unarmed, 3, TRUE)
+		H.mind.adjust_skillrank(/datum/skill/combat/wrestling, 3, TRUE)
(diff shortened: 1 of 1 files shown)


ThePainkiller:
Bit iffy on the stats nerfs, they're supposed to be unarmed specialists after all, and might get a perk to make them unable to wield weapons in the near future, with extra unarmed damage, so the STR should stay as is

Optimism333:
I think the stats are alright, as +2 is generally the level of most upper tier combat classes, and the +1 being bumped to +2 will likely help them a lot with dodging.

I do think their unarmed and wrestling could likely be left at expert (4), however. It's their main gimmick, given they have no armor training and the wooden staff isn't a particularly effective weapon.

CormanQuewye:
> I think the stats are alright, as +2 is generally the level of most upper tier combat classes, and the +1 being bumped to +2 will likely help them a lot with dodging.
> 
> I do think their unarmed and wrestling could likely be left at expert (4), however. It's their main gimmick, given they have no armor training and the wooden staff isn't a particularly effective weapon.

Staff, one-handed has a wdef of 4, if I recall correctly. Meaning it's actually excellent for an unarmed expert to hold in their off hand if they want to parry instead of dodge. Staves are also a classic monk weapon.

1N7Slayer:
i deleted this branch by accident when cleaning my fork up, oops. ill come back to this later

PR #785, opened 2024-09-24
Butchering Update

NPC1314:
## About The Pull Request

What is this?
Mobs files are messy and chaotic to the point of being hard to overview and work in.
Butchering is all over the place with very extreme break points in lethality,, hitpoints and yields.
Lots of smaller details are off, butcher results, what they eat, etc etc.
This is a collection of the various edits and fixes for animals I've been polishing for a while.
butcheroo

This will just improve quality overall, reduce butcher output relative to middling butchers but decrease the lethal threat farm animals presents mostly.

Extremely pointlessly detailed changelog
* Structure
- Restructures the mob files to remove massive amounts of duplication. More organized, easier to interpret the files.
- Adds defines to mob healths. Means you can see all the health levels in one place and also edit them there instead of a file by file basis.

* Feature
- Butchering skill is no longer o-5 speed, 6 double yield.
Now its 0 having 50% to give bad butcher yield
1-5 speed and normal yield
6 gives bonus yield of about 50% so no double chicken
Less nonsensical. Butcher output lowered a bit, if needed the economic value of the animal products might need increasing slightly.
- Gotes now give gotemilk. No practical difference yet beyond taste, its to enable Optimisms desire to make gote cheese items.
and give a little detail for flavor
- Adds fur(pelt( items unique to each animal. No mechanical difference really just looks better. Also updated hide sprite slightly and adjusted sizes of the hide and fur items dropshrink. No more brown colored wool clump for fur.
(Has one drawback, submissionhole is a typecache or such, it works fine, but crafting recipes works on specific paths so requires some duplication of recipes to work, included in the PR. If someone gives weaving any updates I'd recommend grouping pelts into two classes, common (rous and gote) and expensive (volf and mole ) and make them used in different types of quality clothing. Or add recipes that works with the base item and all children but I don't remember how to do that)
- Fixes strange meat not being ingame, (also called spider meat but its just weirdo meat really, subpar.)

* Changes
-Animal headbutts a tiny bit less AP, chickens get a pecking intent thats just crap overall now
-Gotes somewhat less lethal, mostly male gotes get damage reduction, was same as a moo-beast bull
-Chicken slight damage decrease
-All diets updated. Means turnips can be fed to moo-beasts and the like, just updated to be less weird and outdated.
-Slight HP adjustments, saiga lowered a little to be more in line with cows/bulls. Big animal should = big HP.
-All butcher results updated. More or less the same, few of the weirdly poor ones upped (saiga) 
-New emote added, "bows head" that plays instead of "smile" when animals are domesticated. I Think its less weird
that the chickens and saigas bow their head now and than, rather than the little freeks smiling at me.
- Butcher vest was a little too close in hue to the pants, I should have considered contrast a little more when making it. Added slightly reddish-black tint to it, turned out nicely.
- Pig icon gets a tiny hue shift since its contrast was a bit too much compared to the other animals in its pen, blood when dead a little transparent now too. Hard to notice but its slightly better this way.
- TG items removed from butcher results (Trolls and spiders mostly)
- Troll remains placeholder updated with a RT sprite instead of a TG one. Spritesheet lack a remains icon (see mole sheet for example)

Or just keep the current slop and keep the stratmaxxing bros (who inevitably will cry that butcher is now unplayable) happy I guess.

To understand the structure changes the changes below wont help much, just open the files and look and see how a lot was moved to the base rt animal object so inheritance can do its thing properly. 

EDIT
Added the padded gambeson recipe change to lvl 4 and loom since its getting its armor upgraded, as requested by Optimism

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/simple_animal/rogue/game/saiga.dm
+++ b/code/modules/mob/living/simple_animal/rogue/game/saiga.dm
@@ -46,6 +46,7 @@
 	gender = FEMALE
 	footstep_type = FOOTSTEP_MOB_SHOE
 	emote_see = list("looks around.", "chews some leaves.")
+	move_to_delay = 7
 
 	botched_butcher_results = list(/obj/item/reagent_containers/food/snacks/rogue/meat/steak = 1,
 						/obj/item/reagent_containers/food/snacks/fat = 1,
--- a/code/modules/mob/living/simple_animal/rogue/rogueanimals.dm
+++ b/code/modules/mob/living/simple_animal/rogue/rogueanimals.dm
@@ -30,7 +30,7 @@
 	speak_chance = 1
 
 	turns_per_move = 5
-	move_to_delay = 8
+	move_to_delay = 8	// basically speed when player controlled. Lower is faster, a lot faster.
 	see_in_dark = 6
 	robust_searching = TRUE
 
@@ -43,7 +43,6 @@
 	food_type = list(/obj/item/reagent_containers/food/snacks/produce)
 	pooptype = null
 
-	move_to_delay = 5
 	d_intent = INTENT_DODGE
 	minbodytemp = 180
 	lose_patience_timeout = 150
@@ -73,6 +72,9 @@
 	var/food_max = 50
 	var/deaggroprob = 10
 	var/eat_forever
+
+
+
 
 
 /mob/living/simple_animal/hostile/retaliate/rogue/Move()
--- a/modular/Creechers/code/trufflepig.dm
+++ b/modular/Creechers/code/trufflepig.dm
@@ -92,7 +92,6 @@
 	speak_chance = 1
 	turns_per_move = 5
 	see_in_dark = 6
-	move_to_delay = 7
 	animal_species = /mob/living/simple_animal/hostile/retaliate/rogue/trufflepig
 	botched_butcher_results = list(/obj/item/reagent_containers/food/snacks/rogue/meat/fatty = 3,
 									/obj/item/reagent_containers/food/snacks/fat = 1,


lavillastrangiato:
Impressive, very based.

Coastfront:
An unwrapped handle is to snap when whittled; Poacher bandit might need a slight-skill adjustment under this new butchering skill system.

Current Poacher;
0 butchering by default
Gains +1 butchering if old.

If your changes are merged; This would mean that the bandit type that is explicitly-an illegal hunter is somehow shit at butchering animals that he has illegally hunted.

I would thus recommend giving-poacher bandits 1 butchering for free, butchering 2 if they're old probably isn't game-breaking, & the +1 from age can be removed regardless if it somehow becomes a problem.

Additionally, a page found only from seeping ink is bound to have its neighbors stained the same; I would recommend looking through the various jobs that rely on/thematically/Logically SHOULD be at least competent in butchering, as I only found this by virtue of me playing bandit frequently, & My looking over its code for a planned minor-bugfix soon ish.

code/modules/roguetown/roguecrafting/leather.dm
```diff
@@ -47,10 +47,23 @@
 /datum/crafting_recipe/roguetown/leather/heavygloves
 	name = "heavy leather gloves"
 	result = /obj/item/clothing/gloves/roguetown/angle
-	reqs = list(/obj/item/natural/hide = 1, /obj/item/natural/fur = 1)
+	reqs = list(/obj/item/natural/hide = 1,
+				/obj/item/natural/fur/gote = 1)
 	sellprice = 20
 	craftdiff = 1
 
+/datum/crafting_recipe/roguetown/leather/heavygloves/two
```
ThePainkiller, line 55: There's literally no need to do these to clutter the crafting system, recipes always take paths, and SHOULD work with any subtype of fur

NPC1314: Well they don't, so theres that minor snag. So there is actually a very good reason do it like this.

NPC1314:
Ok so after having found the var that toggles subtypes in crafting recipes, its been added, comments explaining how it works.
The leather & weaving recipe pages sorted and labeled better too, the sellprice var in the recipe doesnt work, it needs to be on the actual item so cleaned it up (was only in leather file)

Should sort everything, crafting tests works fine. What will need adjustments during testing migth be yields but its good for a test now.

Optimism333 (approved):
Let's get this tested and ironed out, then... though I'm penciling in a grudge for taking away the double chicken magic.

NPC1314:
double chicken will return as a Dendor spell, have no fear.  (it wont)

code/modules/mob/living/simple_animal/rogue/creacher/mole.dm
```diff
@@ -57,7 +68,7 @@
 	maxHealth = 400
 	butcher_results = list(/obj/item/reagent_containers/food/snacks/rogue/meat/steak = 4,
 						/obj/item/natural/hide = 3,
-						/obj/item/natural/fur = 2)
+						/obj/item/natural/fur/mole = 2)
```
ThePainkiller, line 71: I learned through this PR that someone COMMENTED OUT THE GREATER MOLE? WHY?!

guayitoo:
Good work, improvements to farming and cattle-rearing is the basis of all happy lifestyles.

NPC1314:
> Good work, improvements to farming and cattle-rearing is the basis of all happy lifestyles.

Guayo, ON GITHUB?? The world is upside down, topsy turvy, dogs are walking their owners.
But I do agree, fun farming and ranching will protect us, just tell them to touch the cow  if in doubt.

PR #527, opened 2024-08-15
Cleric Changes and Rebalance (Stats, Armor Training, Weapon Skills)

Optimism333:
## About The Pull Request
Clerics have needed a good and hard rebalancing for a LONG while. I finally got around to just doing it myself.

Unlike my changes to warrior, this could definitely classify as an overall nerf, but it was very much deserved, and not all of the changes were nerfs.

### Significant Changes ###
- Their mace skill has been reduced to skilled instead of expert, though it can still be boosted to expert if playing an old cleric.
- Their armor training has been changed to medium instead of heavy. None of the armor they start with is heavy, and heavy armor training should overall be rarer in my opinion.
- Their bonuses to Intelligence, Constitution, and Endurace have each been reduced by 1. Instead of 2, 2, and 3 it is now 1, 1, and 2.
- Their -2 to Perception has been removed.
- Like templar, they get a psicross AND surcoat based on their chosen diety (in most cases).
- They now also start with a gorget.
- They start with a satchel and a poor pouch of coins.
- Gave their description a little retouch: "Clerics are wandering warriors of the Gods, drawn from the ranks of temple acolytes who demonstrated martial talent. Protected by armor and zeal, they are a force to be reckoned with."
- With clerics getting nerfed in this pr, and their devotion being potentially capped in another, I bumped their available slots from 2 to 4

Screenshot (247)

## Why It's Good For The Game
- Gives a look at rebalancing what was likely the most unbalanced common combat drifter role, and brings it up to speed with some of the new religious customization.

## Pre-Merge Checklist

- [x] You tested this on a local server.
- [x] This code did not runtime during testing.
- [x] You documented all of your changes.

PR #787, opened 2024-09-24
Changes Max Slots for Each Combat Drifter to 2

SaintBurgher:
## About The Pull Request
There has been a long discussion in the Noble's Court (Player Focus Group) regarding the amount of combat drifters that stream into town. Instead of placing a hard limiter on the amount of drifters total that can spawn, it was decided that limiting the combat drifter types to 2 per role would be a better solution.

## Why It's Good For The Game
As it stands currently, combat drifters can endlessly spawn; devaluing the garrison and in-town mercs. This will hopefully curb people endlessly spawning just to jump into combat and die. Effectively giving some OOC motivations to staying alive and not wasting a slot. It will also encourage more people to play town roles if they wish to partake in the ongoing round combat.

Limiting the combat drifter roles and not the "pilgrim" roles is intentional. We still want people to filter into the town, we just don't want the people coming in to devalue the guard, garrison, or mercs.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/warrior.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/warrior.dm
@@ -15,6 +15,7 @@
 	outfit = /datum/outfit/job/roguetown/adventurer/sfighter
 	category_tags = list(CTAG_ADVENTURER)
 	min_pq = 0
+	maximum_possible_slots = 2
 
 
 /datum/outfit/job/roguetown/adventurer/sfighter/pre_equip(mob/living/carbon/human/H)


ThePainkiller:
I love destroying our pop and literally outright ignoring concessions to make it so that towner roles "can join in combat"!

I'm going to be straightforward and cast my vote as junior maintainer to say no to this PR, and to point out a couple more things:

- If this PR comes to comes to pass, people will be utterly mad they can't play their specific desired class
- If this PR comes to pass, Rare classes will be pointless and their balance will be completely thrown off the window, as 50% of the balance decisions surrounding rare pilgrim classes is based on the fact they are **ONE SLOT ONLY** (and the other half being that they have an actual pickchance to roll for them)
- If this PR comes to pass, I will, **effective immediately with no concession nor bargaining**, lower back every non combat role weapon skill down to 1 or nonexistance, rolling back the consession made with towner players after a long and heated debate, to balance out the severe issues in overall class balance this slot limit would bring

I understand the intention, and yes we have to make it so that towner is a more attractive role to play than a drifter with complete freedom enjoying the PVE aspect of the game over the RP aspect of it. But this is definitely not only **not** the way to go about it, it will also throw rare class balance askew.

SaintBurgher:
Currently Drifters are the only role without a hard limit. Why are we limited to a set amount of Mercs/Garrison but a random player can respawn as a combat drifter over and over again to hop right back into the action?

Optimism333 (commented):
Given that barely a month ago the slot limit of 4 on clerics was approved in #527, and the coding team has largely been in agreement to not go in the direction of capping common combat drifters, I wouldn't feel right supporting this.

The direction we wanted to go was a cap of 2 on VERY rare combat drifters, and then see if that's a positive change before making further adjustments.

The lowest slot limit that I would tolerate as a compromise is 4, mirroring the one already in place on clerics. That still wouldn't leave anyone happy, but is much less punishing than 2.

CormanQuewye:
> Currently Drifters are the only role without a hard limit. Why are we limited to a set amount of Mercs/Garrison but a random player can respawn as a combat drifter over and over again to hop right back into the action?

It's an RP server. There is already an incentive to not want to die: you're playing a character and you want the opportunity to roleplay with other characters. If you're wondering: 'if I can endlessly respawn, what's the incentive to not die?' I think this might be a 'you' problem, not a server problem.

ThePainkiller:
I agree with Opti on this one, 4 would be the most we'd like to limit combat drifter slots, the same amount that clerics and mercs have. Anything below that, and we might as well be thinning our pop by trying to force people to play as something they don't want to play. I trust it's not hard to see why this is generally a bad idea.

CormanQuewye:
> I agree with Opti on this one, 4 would be the most we'd like to limit combat drifter slots, the same amount that clerics and mercs have. Anything below that, and we might as well be thinning our pop by trying to force people to play as something they don't want to play. I trust it's not hard to see why this is generally a bad idea.

Right, but at that point, why not limit pilgrims? If you're playing a character who has their background tied to a profession, as (let's say garrison) you know if it's available from latejoin, on ready you have the 'return to lobby' option. On Drifter, you only know what you have after you select it.

You're playing John Rogue? Too bad. You're late to the party. So that plus the fact that it's a limitation that has no reasonable argument for makes me question what would be the point besides screwing people over for daring to actually be invested in their characters', their history, and roleplay?

ThePainkiller:
>If you're playing a character who has their background tied to a profession, as (let's say garrison) you know if it's available from latejoin, on ready you have the 'return to lobby' option, in case you don't get it. On Drifter, you only know what you have after you select it.
>You're playing John Rogue? Too bad. You're late to the party. Guess John Rogue, known for his quick wits and kleptomania is... for this round, inexplicably a barbarian... because Rockhill said 'sorry, bud, four other guys wearing black cloaks already came through'. So that plus the fact that it's a limitation that has no reasonable argument for makes me question what would be the point besides screwing people over for daring to actually be invested in their characters', their history, and roleplay?

Solid points, I was just trying to find a middle ground but honestly I do not agree with this PR entirely. Limiting any kind of drifter is sure to kill pop, and cause frustration down the line.

PR #794, opened 2024-09-25
95%test

NPC1314:
## About The Pull Request

- Adds a 1 in 20 limiter to defense and crafting. 

This means you cannot go above 95% to dodge or parry, and you will never go below 5% to craft. If you have zero chance to craft now it just tells you so and ends the loop.

- 20% malus to defense when laying down added

Less waste of time rolling for crafting, and the chaos of battle is better represented by a 1 in 20 chance to fail.
it gives a slight chance to players whose characters have lower skill level vs higher ones. John Farmer gets a slightly higher chance to hit his victimizer. Better than 1 in a effing 100.
I trust more in general game design concepts inspired by timeless classics such as GURPS or DnD or any of the other 100 RPGs there are over Zeths maths.

Note this needs to be testmerged since its difficult to test pvp alone, crafting bit seems to work fine and didnt notice any issues with dodge when hitting soulless dummies but your mileage may vary.

Laying down idea borrowed from https://github.com/Blackstone-SS13/BLACKSTONE/pull/1215 by ChungusGamer666
Crafting setup is original idea.

## Pre-Merge Checklist


1N7Slayer:
HAHAHAHA ANOTHER MACE BUFF, BASED

Mogelix:
I would prefer taking forever to craft and regretting trying to playing priest and being unable to craft, for instance, a loom to make new robes for acolytes. This crafting change would just close off the opportunities for roleplay opened by the crafting system. Who gains from that exactly? It's not like we have an abudance of people playing seamstresses or crafting roles other than smith.

NPC1314:
> I would prefer taking forever to craft and regretting trying to playing priest and being unable to craft, for instance, a loom to make new robes for acolytes. This crafting change would just close off the opportunities for roleplay opened by the crafting system. Who gains from that exactly? It's not like we have an abudance of people playing seamstresses or crafting roles other than smith.

You don't understand, that comment makes no sense, please re-read the code/description.

What this does is if your chance is 0% it just says so and end the spam.
If its between 1 and 5 % you get 5% chance to succeed to cut down on the tedium.

If anything it does the reverse of what you fear.

code/datums/components/crafting/crafting.dm
```diff
@@ -248,16 +248,20 @@
...
+						to_chat(user, "<span class='danger'>I lack the skills for this...</span>")
+						return
+					else
+						prob2craft = CLAMP(prob2craft, 5, 99)
+						if(prob(prob2fail))
+							to_chat(user, "<span class='danger'>MISTAKE! I've failed to craft [R.name]!</span>")
+							continue
+						if(!prob(prob2craft))
+							if(user.client?.prefs.showrolls)
+								to_chat(user, "<span class='danger'>I've failed to craft [R.name]... [prob2craft]%</span>")
```
ThePainkiller, line 261:
```suggestion
								to_chat(user, "<span class='danger'>I've failed to craft [R.name]. (Success chance: [prob2craft]%)</span>")
```

ThePainkiller (changes requested):

SilvernKell:
Ok, good idea on the crafting QoL, but why the FUCK. Did you slip in a defence malus to being knocked prone? Being knocked prone is already a massive problem and debuff, and a sound strategical choice to making people unable to get away. There was already an attempt to make it less of a game ender with a PR that made it not drop your weapons, which hasnt been pushed yet. 

Yet you're immediately pushing a 20% defence malus to being knocked prone. If I'm being uncharitable I'm saying is this an attempt to buff anthraxi whips or something?
Please don't do this. This is a really bad idea. Un-fun weapons do not need to be buffed and it doesn't feel good to be defeated by them.

or on the other hand, do it, and i'll laugh as archer gets stronger, polearm gets stronger, running away and not fighting in close quarters gets stronger. Let's watch everyone being impossible to protect as people switch to pelting people wordlessly from above as it's the only way to deal with bandits and antagonists.

ThePainkiller:
It's a -20% less **chance to parry** while you're prone. It makes sense that lying down doesn't allow you to defend yourself properly. It is also good for maces, it's not to buff the whips.

1N7Slayer:
It is never a sound strategic choice to be knocked flat on your ass and scrambling to crawl away, because if that does happen in a medieval combat scenario, a couple of things could and should happen to you:

1. A burly knight should drop his sword, drop himself on top of you, and start stabbing you to death in the heart, throat or eyes with his dagger.
2. A murdering soldier should just bash your brains out with his mace,

1N7Slayer:
> There was already an attempt to make it less of a game ender with a PR that made it not drop your weapons, which hasnt been pushed yet.

Not true, it was pushed and is currently in game now.

ThePainkiller (approved):

PR #812, opened 2024-09-27
Bandits now have 0 PQ requirement.

1N7Slayer:
## About The Pull Request

Bandits (an antagonist role with zero to negative responsibility and a lot of power/potential/stealth), did not have PQ requirements for some reason. Now they ideally will.

## Why It's Good For The Game

It's the only late-join antag role in the game, kinda hard to parse for whether there are bandits in the round unless you pay attention to bandit static names. I think it is best we gloss over this stuff collectively and admit this is a role we want players who get along well with others in order to be this kind of antag causing trouble in a fun way for others without resorting to permabanning players in the future, and instead relying on PQ adjustments to meter if someone is good for a specific set of roles without restricting too much their ability to play here.

I will say as a result of this PR, I think staff should be more inclined to give people PQ resets from this point on, and future antag coders need to reference this PR as to why we might need PQ floors on antag roles. This one is the most pertinent for the time being.

## Pre-Merge Checklist


Optimism333 (approved):
Simple and reasonable change.

ThePainkiller (approved):
Yeah, I don't see why not, let more play the stock bad guys

PR #825, opened 2024-09-30
Soilson Love

NPC1314:
## About The Pull Request

Small collection of soilson related fixes and tweaks

* Thresher had a annoying sound reverb when threshing, gone now
* Walking on seeds and planted plants negative effect lowered even more
* Thresher now backslot too and thus uses a normal sized sprite now. Its just necessary for all tools 
to be backslottable and normal sized sprites, unlike fighter classes who might be able to handle one cumbersome main weapon if good enough, these tools are not optional,. you must use several of them and they all are not powerful at all.
Just have to be this way, simply put.
* Fyritius flowe was TG hydro code still, refactored to new standard. New art added, based on old. Can now be grown.
* Sound effect from interacting with grass/bushes/produce normalized and softened a little, was very loud and sharp
* Kinder rot times for swampweed and westleaf
* Turnip yields raised a little, single slice results makes this needed. Also made into cash crop just like wheat.
Only useful for soups and animal feed ingame, but got very slightly edge over wheat in yield per time unit so functionally a minor bump up in farmer income
* Sellprices of all produce set to 1 as floor. Previously was 0. Produce should have economic value, 1 mammon = 1 fruit or veggie is as good a measure stick as any.

None of this is very important but it adds up. TLDR thresher on back, turnip good, sell turnip.

## Pre-Merge Checklist


ThePainkiller:
wdym it's not very important, these are amazing QoL changes

Issue #365, opened 2024-07-12
[Suggestion]: Make Alcohol Bottles Hold A Bit More Liquid ~24 oz Instead of 15 oz


Optimism333:
### Category

Balance

### What is the suggestion?

Simply make wine bottles hold a little more liquor, so you can pour 3 full mugs/goblets rather than barely fill 2.

As a small addendum to the suggestion: 
Bottles of different sizes? A smaller bottle (beer) holding 16 oz, and a larger one (wine) holding 24 oz.

### Why is this suggestion good?

A little more convenience for servants pulling wine up from the cellar. Two wine bottles would serve six guests at a feast, or eight if the amount poured was 6 oz per cup. It would allow serving more drinks to a depressed noble from a single bottle: 3 full goblets instead of barely 2.

More alcoholism.

Makes it easier to be the liquor.

### Staff consulted

N/A

Mojamble:
The merchants guild supports this.

cornercube:
quick and easy fix, I will get to it

cornercube:
![image](https://github.com/user-attachments/assets/1e3c8013-e699-4797-9e7f-8ad63f22da95)

could start with a global 25 unit buff
small and magnum bottles would need to wait for their own sprites

PR #833, opened 2024-10-01
big gulp: stonekeep

cornercube:
## About The Pull Request
increases base bottle capacity to 70cl or ~24oz (23.65oz)
merchant wines buffed to have 70cl inside them as well, so you get more booze for your buck when ordering it 

## Why It's Good For The Game
#365 

already merge worthy
though it would be nice to have 1500ml large magnum bottle sprites, to allow the king to truly indulge

## Pre-Merge Checklist


AceVX:
Awesome. No more one and a half cups of wine per bottle

Optimism333 (commented):
I've wanted this for a long time given how quickly wine bottles run out. The servant powercreep is real.

Will wait for a while though to see if there's anyone who has opposing feedback.

ThePainkiller (approved):
No qualms about this, except that maybe we need to make gulps take even units of liquid rather than 3u at a time

PR #849, opened 2024-10-02
Poison tweaks

NPC1314:
## About The Pull Request

Suggest testmerge over couple days and tweak toxloss if needed

Second tweak of poisons post alchemy changes

Poison been a little weak, most likely the addition of
	if(volume > 0.99)
post testing affected it quite a bit. Then theres a possiblity tickrate on server dragging it out even more so might need to slightly adjust toxloss values too, but lets start here. Realized this actually halves damage from poison arrows. 
1 unit cutoff is just far too big with doses of 2-3 u being most common.

This all relies on gulp sizes of 5 btw
Tweaks phase 1
	if(volume > 0.09)
Tweaks phase 2 (if more is really needed, should test phase 1 for at least couple days)
 bump up toxloss by 0.5 per cycle, evaluate for a week.
 
Killersice or other instant kill poisons has no place IMO, disabling (berrypoison) and killing (strong poison) is enough to balance and both tie in nicely with alchemist  and their antidote potions, giving a reason for the antidote to exist, for alchemist to make it, for king to interact with alchemist, gives the target player a little window of time to act if poisoned and have prepped right and is on his toes.

## Pre-Merge Checklist


ThePainkiller:
The volume limitation on poisons was so that people stopped using 1 unit of killersice dilluted on an entire barrel of water and dilluted the poison as much as they could. With this, poison is back to its lethality even at 0.0001u inside someone's body.

Optimism333 (approved):
Good changes, noted in-game that blindly testing berries will take a chunk out of your health again rather than do nothing.

PR #851, opened 2024-10-02
Hand Buffs (Tis peak)

BadAtThisGame302:
## About The Pull Request

Gives the so called Spymaster more Rogue skills... like SNEAKING.... one extra climbing and one extra on crossbows.

Also gives him the ability to see prices of items, nudging him towards the Steward/Courtier angle aswell.

Nerfs his INT/PER by one and gives those two points lost to his STR/CON

Gives him knives 4 if he's old

And removes the unused walls key and replaces it with a Merchant key

## Why It's Good For The Game

The so called and constantly shouted at Spymaster when even a beggar has more sneaking than him... tis peak!! Not to mention the see prices fluff and the extra crossbow/climbing for that Royal Rogue vibe people keep screaming about.

## Pre-Merge Checklist


code/modules/jobs/job_types/roguetown/nobility/hand.dm
```diff
@@ -43,20 +43,26 @@
 	backr = /obj/item/storage/backpack/rogue/satchel
 	backpack_contents = list(/obj/item/rogueweapon/huntingknife/idagger/steel = 1, /obj/item/keyring/hand = 1)
 	if(H.mind)
-		H.mind.adjust_skillrank(/datum/skill/combat/crossbows, 2, TRUE)
+		H.mind.adjust_skillrank(/datum/skill/combat/crossbows, 3, TRUE)
```
ThePainkiller, line 46:
```suggestion
		H.mind.adjust_skillrank(/datum/skill/combat/crossbows, 2, TRUE)
```
Not a combat unit, no.

ThePainkiller (changes requested):

Optimism333 (commented):
Role might have been a bit overly nerfed. I think it will need a more thorough look at some point... but this is fine in the meantime. 

Will give it a day to see if there's any more objections or feedback.

NPC1314:
Sneaky hand makes sense but I will say this, stats on roles keep changing and theres no system to it.
The old rule of thumb gives the statline as +9 post these changes, from +6 before (str and speed count as 2)

Is that reasonable? I dont know.

Optimism333 (approved):
No blatant opposition, even after a couple days.

Stats are a bit high, but not insane for a single-slot role that is supposed to be the King's best buddy. Let's see how this goes.

PR #852, opened 2024-10-02
Map Changes: Moves Veteran's House and Adds in Rockhill Lake

SaintBurgher:
## About The Pull Request
Moves the Veteran's House to be inside the Southern Wall. Replaces their old house with a small pond added guardsman watch post. Also fixes the village gatepost being dark as well as some railing fixes for the garrison checkpoint and Southern Gate checkpoint.

## Why It's Good For The Game
The Veteran is in a weird spot where they are very far away from where they need to be (the village). This positions them in an area that will show their importance to not only the village, but the garrison. Placing a nice pond where their home used to be will hopefully give that side of the map some more action. This also gives garrison members a new vantage point (or area to go and get drunk). All-in-all, some fluff changes that will hopefully help facilitate roleplay.

Veteran's House
373789962-0dcd4164-297b-44ad-8c98-7ed7af52b4e5
image

Pond and Watch Post (Has Stone murderholes facing North)
image
image

Railing Fixes
image
image

## Pre-Merge Checklist

### DIFF
--- a/_maps/map_files/dun_manor/dun_manor.dmm
+++ b/_maps/map_files/dun_manor/dun_manor.dmm
@@ -176,15 +176,12 @@
 /turf/open/floor/rogue/cobble,
 /area/rogue/under/town/basement)
 "akh" = (
-/obj/structure/mineral_door/wood{
-	locked = 1;
-	lockid = "garrison";
-	name = "veterans house"
+/obj/structure/fluff/railing/border{
+	dir = 8;
+	icon_state = "border"
 	},
-/turf/open/floor/rogue/ruinedwood{
-	icon_state = "weird1"
-	},
-/area/rogue/indoors/town)
+/turf/open/floor/rogue/cobblerock,
+/area/rogue/outdoors/town)
 "akL" = (
 /obj/structure/fermenting_barrel,
 /turf/open/floor/rogue/ruinedwood/spiral,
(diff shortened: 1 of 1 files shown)


Ickysoup:
People will break down the vets doors to get into town making the double gate useless

ThePainkiller:
> People will break down the vets doors to get into town making the double gate useless

I agree with this, maybe don't give it a door that leads directly to the Mayor's house backyard and such, that can be easily picked, bypassing the entire gate, feels like a huge security flaw

SaintBurgher:
> People will break down the vets doors to get into town making the double gate useless



> > People will break down the vets doors to get into town making the double gate useless
> 
> I agree with this, maybe don't give it a door that leads directly to the Mayor's house backyard and such, that can be easily picked, bypassing the entire gate, feels like a huge security flaw

By the time it takes for them to bash down the two doors they could have already climbed the tree and hopped over the wall though. This still wouldn't be the fastest or most convenient way in.

I can remove one of the doors if we think it's necessary but in the grand scheme of things someone would need to bash down:

1. A palisade
2. Outside door
3. Inside door

All without getting caught or killed. If anything this may make less people take the "easy" way in of just hopping from the roof onto the wall and then climbing down.

NPC1314:
We are prioritzing the TGM test so please hold off on this one until we got that testing done, planned to be done in 12 hours

SaintBurgher:
Final verdict on this? I can make any changes necessary.

SaintBurgher:
New vet house entrances
image

NPC1314:
Ok so we have moved to TGM now, thats the "modern" TG system where the map is basically modular so people should be able to merge maps without issues as long theyre not makking the exacxt same spots.

What this means is this needs converting to TGM too , using StrongDMM for example.

Optimism333 (approved):
Looks good to go!

PR #866, opened 2024-10-04
Deadites Revamp

NPC1314:
## About The Pull Request

Makes Deadites slower, deader, less prone to drinking, better at smashing windows, worse at dumping rocks on heads.

With some fresh bugreports its clear there is more to do. Turns out the new stats only apply when getting deadite by admin command, when dying you get other set so.

* Deadite stats sorted (all get speed 3 and Int 1 now)
* Claw intent get its own button now
* Clawing doors and windows now breaks them rather than knocks
* Stones and sacks can't be picked up by deadites no more to prevent some silly borderline griefy behaviour
* Removes dead people and zombies being nauseous
* The instant door break replaced with heavy damage to doors to give a few seconds warning before deadite bursts in.
Windows now break instantly though.

Also fixes the terribleness of https://github.com/Darkrp-community/OpenKeep/pull/273 by just removing the
osmosis/sponge effect that made you drink the water constantly when wading in it.
* Removes dead people keeping drinking when dead
Water tile fix will remove some silly and reduce workload very slightly

* couple destroyed sounds normalized
* a small folder of TG programming icons not used in any way yeeted, utterly useless and I dont wanna accidentally click it again.

## Pre-Merge Checklist


BadAtThisGame302 (commented):
Should use span_info and span_warning instead of the long span class etc etc...

NPC1314:
> Should use span_info and span_warning instead of the long span class etc etc...

true true, getting a bit lazy

code/game/objects/structures/mineral_doors.dm
```diff
@@ -150,6 +150,15 @@
 				force_open()
 				user.visible_message("<span class='warning'>[user] smashes through [src]!</span>")
 			return
+		if(HAS_TRAIT(user, TRAIT_ROTMAN))
+			if(locked)
+				user.visible_message("<span class='warning'>The deadite bashes into [src]!</span>")
+				take_damage(50, "brute", "melee", 1)
+			else
+				playsound(src, 'sound/combat/hits/onwood/woodimpact (1).ogg', 90)
+				force_open()
```
ThePainkiller, line 159: This is a zero warning one hit instaopen doors if they're locked, what's the point to lock doors if it literally makes you more unsafe?

NPC1314: Its a REDUCTION to 1/4 of what it is currently

NPC1314: Currently they have doorbasher which is 200 damage per bump, same as werevolf

PR #869, opened 2024-10-04
Wound Pain Tweaks

anything86:
## About The Pull Request

Quick Summary: Fixes missing pain values, reduces artery pain to compensate, slightly reduces generic fracture pain.

Slash and puncture wounds, as well as a few other minor wounds (punctured eardrums) were all missing associated pain values. 
This change adds basic pain levels to such wounds, while keeping them firmly underneath bruises in terms of pain inflicted. It also reduces artery wound pain to compensate, since some amount of slashes and stabs are nearly mandatory to inflict those and they now inflict pain by themselves.

Other minor changes include: 
-bringing generic fracture pain more in line with other non-life threatening wounds
-adding a hefty amount of pain to pierced eardrums
-reducing the pain on small bites and reducing the pain for individual eye wounds (when compared to both eyes being gouged) for reasons similar to the ones cited for artery wounds
-small amount of pain added to tongue dissection

## Why It's Good For The Game

The result of the previous status quo was that fights involving any manner of stabbing or slashing weapons would only really end once an artery was severed through a critical hit. This is arguably unrealistic, undermined the value of the pain threshold stat (Endurance) and asserted blunt damage's dominance even further beyond its knockdown utility.
This should bring stabbing and slashing weapons more in line with blunt, and reduce their reliance on critical hits.

## Pre-Merge Checklist

### DIFF
--- a/_maps/map_files/roguetown/otherz/smallforest.dmm
+++ b/_maps/map_files/roguetown/otherz/smallforest.dmm
@@ -61,8 +61,8 @@
 /area/rogue/outdoors/river)
 "aq" = (
 /turf/open/water/river{
-	icon_state = "rockwd";
-	dir = 4
+	dir = 4;
+	icon_state = "rockwd"
 	},
 /area/rogue/outdoors/river)
 "ar" = (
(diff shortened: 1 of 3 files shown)


NPC1314:
Considering the absolute canine droppings we been stuck with so long theres no way this makes it worse.

Values checks out, only "issue" is its not part of a bigger system overhaul/fix but its gonna give good mileage, as bugfix it gets prio for testing.

ThePainkiller:
tbh this is fair, I feel pain never comes into play during combat to stun someone unless they have mace-inflicted wounds so this is a nice change for piling up other kinds of wounds 👍🏻

NPC1314 (approved):

PR #872, opened 2024-10-05
TGM test

NPC1314:
## About The Pull Request

Tests out how TGM merges working out for us while also kind of fixing the issues with adventurers spawning in water
* Adv spawns moved slightly or surrounded by couple more squares of land so the random drop hopefully always puts them on dry land
* couple  weird areas and a saiga walled in sorted
* adv spawns in swamp removed, better get the murderwoods adv spawn route and systems working before just adding more halfbaked.

## Pre-Merge Checklist

### DIFF
--- a/code/game/objects/effects/landmarks.dm
+++ b/code/game/objects/effects/landmarks.dm
@@ -727,7 +727,7 @@ INITIALIZE_IMMEDIATE(/obj/effect/landmark/start/new_player)
 /obj/effect/landmark/underworldsafe // To prevent demons spawn camping will save a lot of ear rape.
 	name = "safe zone"
 
-/obj/effect/landmark/underworldcoin/Initialize(mapload)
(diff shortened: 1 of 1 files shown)


sanshoom:
Now that I think about it, I don't think the adventurer spawns in the bog ever worked when they were re-added.

NPC1314:
Not surprised, just as well tbh since most would have been dumped in water with no sensible path out. Has its charm but shoving every pilgrim in there is a bit much.  For veterans choosing speciifc roles knowing its a challenge it would be fine but we lack the content for it right now.

NPC1314:
Turns out I was in the wrong folder, all that was needed was a bit wider zone of dry land for a couple of the spawn spots, reduced the scope to that.

Optimism333 (approved):
Let's hope spawning in and drowning will soon be a thing of the past, finally.

PR #879, opened 2024-10-06
Bug fix about ore and cleaning floor tweaks

BadAtThisGame302:
## About The Pull Request

- Cleaning floors will no longer spam for every attempt, now only making a message on successful cleaning, and in smaller font. 
- Putting ore in a smelter should appropriate remove the intents of the item from your hand.

## Why It's Good For The Game

Cleaning clogs up the chat, so this is an attempt at negating that.

And the intent missinputs when you don't have ore in your hand are now fixed.

## Pre-Merge Checklist


code/modules/roguetown/roguejobs/blacksmith/smelter.dm
```diff
@@ -55,6 +55,7 @@
 		return ..()
 	if(W.smeltresult)
 		if(ore.len < maxore)
+			user.dropItemToGround(W)
```
ThePainkiller, line 58: Not the most...elegant of fixes, shouldn't there be a proc to update the hands of who we take items from?

ThePainkiller: Nvm I just realized this calls exactly for update hands and icons, it works exactly as intended, LMAO disregard my comment pls.....

ThePainkiller:
Can you please add this fix as well to adding items manually to the handcart? I beg you.

NPC1314 (approved):

PR #903, opened 2024-10-08
New Bushes

NPC1314:
## About The Pull Request

Adds new type of bush  (bush/meagre)
Slight icon adjustment (green subtype added, some clickability fixes, instead of 2 duped ones in the red list one was edited to get mroe unique looks)
It does not work like a vendor with three very likely content items 
Instead its random species, greener could have berries, redder could have tobacco leaves, both might have thorns or fibers
Scavenging them yields results depending on perception and fortune stats, once something has been harvested its depleted, slight visual cue, also obviously when clicked

It will tie in stats a little more, reduce the vendorized feel of bushes we got now, and reduce amount of free food

Also maps in this new bush to most maps and mapgens
Also nerfs berry nutrition and soups, mostly the veg ones
Also adjusts a couple sell prices for produce (dried tobacco & swamp weed got value, raw ones do not, fast rotting stuff not exportable)

Tacked on the discussed bog tweak, bogbugs bugged bigtime so temp disabled for bog ambush, maneaters chance increased in bog specifically, theres fewer tiles for them to spawn so despite already higher than forest already, theyre rarer, this will help a little.

EDIT
New bush added, the bog bush, often home to silk grubs that can be squeezed to eject the silk string withing (a way to find silk without combat/spiders, unique to bog) 
Added 1% to fake maneater spawning in bog too (was no fake ones before)
Fixed a bug where droprates werent resetting, making it harder to find thorns, also adding a good chance 
cutting down bushes will yield a thorn
Fixed thornbushes, can be searched for thorns
Moving through bushes/thorns improved, now has delays and risks of getting thorned depending on move intent.
Icons for the hedge and the wild shrubberies no longer shared, there was enough of them to make two separate categories. Lets you visually identify the cultivated ones better.

Overall this will make hunting and various food strats matter a bit more outside town, increasing value of food a bit.

## Pre-Merge Checklist


code/game/objects/structures/roguetown/rogueflora.dm
```diff
@@ -649,3 +649,85 @@
...
+
+		if(prob(prob2findstuff))
+
+			if(prob(prob2findgoodie))
+				var/obj/item/B = goodie
+				if(B)
+					B = new B(user.loc)
+					user.put_in_hands(B)
+					user.visible_message(span_notice("[user] finds [B] in [src]."))
+					islooted = TRUE
```
ThePainkiller, line 719: Bruh, so bushes will just give ONE item and only ONE item forever, for anyone that picks it first, and that's it? This is a massive nerf.

NPC1314: Think its too harsh? Maybe, kinda the point to reduce the flood of  berries outside town tho, theres a metric ton of bushes around. Fibers are easy to get from grass and theres usually tons of discarded thorns around, not sure its going to be a problem tbh.

Optimism333: Personally I think the big hedges "great bushes" that grow in the gardens of more decorated areas should give more than a single item. Since they're a subtype of bushes, this single item change might trickle down to them, and I'm not sure I like that.

And one other change I would suggest is that if someone has good fortune, maybe have a small chance for the bush to not be be looted after use.

NPC1314: Its a new bushtype, didnt replace any of the tall hedge style ones in this PR, they still as plentiful as before.

I like the fortune idea tho, Ill tinker

Optimism333: Sounds good, then!

Optimism333:
Screenshot (444)

Used bushed probably need to be a darker or more vibrant green. They blend with grass too much, in my opinion.

NPC1314:
> 
> Used bushes probably need to be a darker or more vibrant green. They blend with grass too much, in my opinion.

True, will do a tweak and up the luck modifier a bit for bonus yields

anything86:
Interestingly, the green bushes do not block movement from the south direction as the old (red) bushes do. I always wondered whether it was intentional that the old bushes did that, and it's really annoying when you have them in a corridor pointed south. Either way, might be worth standardizing the behaviour so that neither block movement, or both do.

NPC1314:
> Interestingly, the green bushes do not block movement from the south direction as the old (red) bushes do. I always wondered whether it was intentional that the old bushes did that, and it's really annoying when you have them in a corridor pointed south. Either way, might be worth standardizing the behaviour so that neither block movement, or both do.

Standardization is needed, agreed. The move block from south on old never made sense to me, its a shrubbery, shouldnt be directional, but at the same time its a little too fast to traverse right now, a slowdown is needed.

Currently all the randomly spawned bushes are this new type, and the premapped ones outside of the tall garden ones too, but there might be some exceptions that slipped through the cracks somehow, varedited map stuff for example.

What area did you encounter bushes full blocking movement?

anything86:
> > Interestingly, the green bushes do not block movement from the south direction as the old (red) bushes do. I always wondered whether it was intentional that the old bushes did that, and it's really annoying when you have them in a corridor pointed south. Either way, might be worth standardizing the behaviour so that neither block movement, or both do.
> 
> Standardization is needed, agreed. The move block from south on old never made sense to me, its a shrubbery, shouldnt be directional, but at the same time its a little too fast to traverse right now, a slowdown is needed.
> 
> Currently all the randomly spawned bushes are this new type, and the premapped ones outside of the tall garden ones too, but there might be some exceptions that slipped through the cracks somehow, varedited map stuff for example.
> 
> What area did you encounter bushes full blocking movement?

Out of the top of my head, there is a tight hallway at west Mt. Decap that is always blocked by a red bush, as well as a corner just north of the mining camp north of town where you can get stuck as well, between the palisade, trees and a red bush. I've died once as a result of the later, since there is not much you can do if cornered in these places (jumping over the bush is not possible). This happened during the testmerge of this PR.

On a semi-related note, great bushes (manor) block movement on all directions (not only south), which includes if you happen to drop on top of one by climbing, although great bushes are a lot more rare and probably not an issue.

assortedbeads:
This has made finding thorns almost impossible, Ive had to go through 10 bushes of smokeleaf and berries just to find a single one.

Someone mentioned its cause its gated behind failed chances? Idk but finding a thorn shouldnt be harder than finding literal berries

guayitoo:
Hey there would you please add a quick bush specific to the Bog (if not possible that's fine)
I'd like these big bush looking trolls to be able to blend into their environment if possible!

We'd need to edit the bog troll squatting sprite to make the feet look like protruding roots.

@NPC1314 
image

image

NPC1314:
Thats perfect!

The red bushes that behave differently will get gatted.
I'll take another look at the maths, original plan was about 40% chance to get a thorn so maybe somethings off if you getting that few.
The bog bush is perfect though, a new thorn bush just for bog which mainly yields thorns (and trolls)

YuiY1997:
Thank you for the thorn drop chance increase

NPC1314:
There were some issues with the math but thats sorted too now, more or less behaves as designed now. Next round should have the correct version.

Optimism333 (approved):
Main issue I heard of was thorns being difficult to acquire. Looks like that's been addressed, and this is has been in testing for a while. Time to put it in the pipe.

PR #906, opened 2024-10-08
Toll timer and handcart intent fix

NPC1314:
## About The Pull Request

The underworld is dire here. Never actually died before, well, never went to the underworld, it blows.
Adds a timer, copied from Blackstones code, that spawns a toll in your hand 10 minutes after you travel to the underworld
So no need to rely on mapped coins, 10 min respawn timer seem endurable.

Also uses the same solution as #879 to fix intent muck when loading handcarts.

## Pre-Merge Checklist


Optimism333 (approved):
Given the recent focus on the underworld due to its jank... better to get this qol in sooner rather than later.

PR #915, opened 2024-10-09
Minor Map Fixes

SaintBurgher:
## About The Pull Request

- Fixes the house south of Feldsher meshing in with the walls/roof
- Fixes the cobble up near the new towner houses to not be excessive
- Allows the stable to be unlocked by the house key next to it (house II)
- Standardizes roman numerals on houses to be uppercase

## Why It's Good For The Game
More map cohesion.

**HOUSE FIX**
image
image

**COBBLE FIX**
image

**SIGNS**
image

**MAYOR'S TABLE**
image

## Pre-Merge Checklist


ThePainkiller:
imagen

What I'd suggest is to nix that entire upper part of the wall there, lower the ladder one square to the south, and then eliminate that stupid cuck chair that makes no sense, and take the window down south as well so that it makes sense in the middle of the room

Do that on the bottom floor as well, and you can have ceiling-type flooring connecting this house and the feldsher's with no issue so that rogues can escape through these (otherwise those windows don't have a point existing)

SaintBurgher:
> imagen
> 
> What I'd suggest is to nix that entire upper part of the wall there, lower the ladder one square to the south, and then eliminate that stupid cuck chair that makes no sense, and take the window down south as well so that it makes sense in the middle of the room
> 
> Do that on the bottom floor as well, and you can have ceiling-type flooring connecting this house and the feldsher's with no issue so that rogues can escape through these (otherwise those windows don't have a point existing)

image
image

Optimism333 (approved):
Looks good to go.

PR #918, opened 2024-10-10
Blunt Knockdown Fix

anything86:
## About The Pull Request

The formula for resisting blunt weapon knockdown had an oversight in its implementation, resulting in comically low chances even at legendary athletics. This addresses this issue.

This also serves to gather feedback on possible alterations to the formula, but for now it'll only be fixing what is obvious.
For reference, at 10 speed, skilled athletics, you had a 88% chance to be knocked down previously, this fix takes that down to 61%. Both assume a knockback distance of 1 tile, as is most common.

## Why It's Good For The Game

Better balances blunt weapons, makes athletics skill more relevant. It was a bizarre but accepted fact of life before that a light slap with a cudgel onto plate armor would almost always knock you on your ass.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -2874,8 +2874,7 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 	var/skill_modifier = 10
 	if(istype(starting_turf) && !QDELETED(starting_turf))
 		distance = get_dist(starting_turf, src)
-	if(mind)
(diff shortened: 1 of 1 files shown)


anything86:
Moving to review. Personally feel as if other factors of the formula could be adjusted to result in an overall lower knockdown chance, as I feel 61% for an average fighter is still high, but the primary intention is just to fix a bug.

NPC1314:
This wasn't actually testmerged, did so now.

NPC1314 (approved):
We roll

PR #920, opened 2024-10-10
Changes Gambeson Default to be more Grayscale

SaintBurgher:
## About The Pull Request
Changes gambeson default sprite to be more grayscale so that they can be dyed better with less color overlap.

## Why It's Good For The Game
Dying gambesons at the moment applies the dye over an already light brown sprite. It makes the results muddled and they don't look as clean. This will fix that!

**DYED**
image

## Pre-Merge Checklist


SaintBurgher:
image

ThePainkiller:
What happens to people who just want the plain brown gambeson as usual?

NPC1314:
> What happens to people who just want the plain brown gambeson as usual?

In theory you could add something similar to the  default  brown shade to the initial item, but it looks a lot like some shading depth was lost as well in the transition to grayscale.

The colored gambs looks very saturated, an issue to some extent with all dyes due to the selected hues, but unless the default one looks just as good as the current one when you spawn with it this one is not worth the cost.

NPC1314:
I just did a quick full desaturation of the old gamb and it looked like this.
desature

Yours look like this
desaturefail

They will never look as good again no matter what color due to lost shading.

PR #926, opened 2024-10-11
Fixes Legendary Peasant Missing Gambeson (Bugfix)

Optimism333:
## About The Pull Request
- Evidently the gambeson can't be worn with a work vest, even as a shirt.
- Removed legendary peasant work vest, shifted their gambeson spawn to the armor slot and gave them an undershirt

## Why It's Good For The Game
Fixes gear not spawning properly.

## Pre-Merge Checklist


ThePainkiller (commented):
As someone found out from way back, this happens because both items share the same drop item sound. I know it's dumb, but that's how nonstacking armor is handled. I suggest instead giving only gambesons their own unique sound so they don't overlap with other clothing items, and can be worn with things like vests, etc.

NPC1314:
That sounds rather strange, and a quick test doesn't seem to support the sound theory. At least a leather vest can be worn under a gambeson despite having the same dropsound. What proc would be causing this?

NPC1314 (approved):

PR #937, opened 2024-10-13
Blood & Tox heal tweak and minifixes

NPC1314:
## About The Pull Request

* Lowered heavysmash misscost further to 23 from 25. Reason: So you can actually make 4 swings before falling down and dying.
* Fancy var added to bottles so you can choose if you want classic style or keep the desc when uncorking etc. Best of both worlds.
* Cyanide bottles mapped in replaced with strong poison ones (the poison actually balanced for RT)
* Sleepheal rates vs poison halved.  Before it was up to 3!! per tick, a crazy amount
* Waters and soups bloodloss replenishement reduced from 10 to 4 and 6 respectively. Water being x3 as effective as straight blood transfusion is kinda derpy, now its same as the bloodloss regen from sleep, combined for double. Soup the best blood medicine until proper alchemical or feldsher stuff added, suffices.

* Ghosts can now pass through most doors, need to be next to them and click them
* The coin upkeep proc commented out, tiny blow against lag, but its not needed since timer for Toll and maze thing soon.
* New prompt in chat in green when a player dies, giving the tip that they need to Journey to Underworld and just wait a little for a respawn 
* Perfume spam fixed

* Tweaked some things with crops, simplified the way the crops are categorized
* Finally does step 2 of the meta-hunting minor helper of key names. Its very marginal but default key name is now "Old key" and its used for antags as well as every non-town key in dungeons etc, as well as for both walls key and crypt key. The desc on both helps clue you in to their purpose. Just to make it a little harder to tell at a glance at inventory theyre a baddie.

Found out fermenting barrels were kinda silly, fixed. Added small sound for loading them.
Adjusted amounts of liquior you get, expensive strong stuff like cognac is smaller amounts than ale ofc.
Most important, you'll now see the real color of the contents! neat huh? Adjusted the filler a little, was very transparent
Also  yeeted some random TG food and bottle items not useful that I stumbled over.

## Pre-Merge Checklist


Optimism333 (approved):
Lot of good qol and fixes, it's been in testing for a while, let's get it in.

PR #947, opened 2024-10-14
You can no longer sleep in armor

BadAtThisGame302:
## About The Pull Request

Ports this wonderful Pull Request from our very kind friends over at Ratwood, yippie! Here's the PR https://github.com/Rotwood-Vale/Ratwood-Keep/pull/758 and the author was kind enough to comment, wowza. Give @GuppyLaxx a wave and a hello in the comments for his awesome work!

Most armors I found fitting to block are blocked, all heavy ones and some medium ones, helmets vary from light to heavy and such. I am sure you'll codedive and find out.

## Why It's Good For The Game

Because I have had enough of people thinking they're safe ordering plate armor and never ever taking it off even when they sleep. You're not superman, you would in any realistic scenario have died inside the armor if you kept it on as long as the average Stonekeep player.

## Pre-Merge Checklist


sanshoom:
based

ThePainkiller:
What about making sleepytime a thing based on last time you fell asleep rather than tying it specifically to global night time instead?
With how sleepy debuff works right now this is gonna do more harm than good

Technobug14:
This seems like a really odd change. Having to take off all of your armor, especially now that I believe more armors were given do_afters to take off and put them on, this just seems like a time waster for no real reason? I mean, if the idea here is that it's easier then to kill someone in their sleep, just take the armor off and stab them? They're asleep. They're not going to swat your hand away and prevent you from removing it. This is just adding tedium for the sake of tedium, the only thing that could be stated is realism or immersion and I don't think the status quo is that immersion breaking.

NPC1314:
Perhaps a less radical version would be suitable?
I 100% think its a improvement to have this flag on mask/head slot items, armor makes sense but its a little tedious to remove and put on all that stuff. 
Maybe just do head/mask items in first go, evaluate?

Suggestion is then to:
Add flag to 
* most items in head, mask slots  (helmets and hats and such, just slap the flag on the root headwear and mask items I guess, ez)

And hold there., test, see whats what.

BadAtThisGame302:
> Perhaps a less radical version would be suitable?
> 
> I 100% think its a improvement to have this flag on mask/head slot items, armor makes sense but its a little tedious to remove and put on all that stuff. 
> 
> Maybe just do head/mask items in first go, evaluate?
> 
> 
> 
> Suggestion is then to:
> 
> Add flag to 
> 
> * most items in head, mask slots  (helmets and hats and such, just slap the flag on the root headwear and mask items I guess, ez)
> 
> 
> 
> And hold there., test, see whats what.

A rare amount of people wear masks and usually they only wear them to fight. And I don't find it fitting not being able to sleep in some current headgear thus why not all of them are barred from such. The issue is not even in the masks and helmets it's in the fact that nobody ever takes their armor apart to heal their wounds. This is the whole intention of the PR, to be radical. Not to add the var to the non issues 99% of the time like helmets and masks.

BadAtThisGame302:
> What about making sleepytime a thing based on last time you fell asleep rather than tying it specifically to global night time instead?
> 
> With how sleepy debuff works right now this is gonna do more harm than good

Night based sleeping is better and makes more sense imo so I won't be the one to change that.

anything86:
> This seems like a really odd change. Having to take off all of your armor, especially now that I believe more armors were given do_afters to take off and put them on, this just seems like a time waster for no real reason? I mean, if the idea here is that it's easier then to kill someone in their sleep, just take the armor off and stab them? They're asleep. They're not going to swat your hand away and prevent you from removing it. This is just adding tedium for the sake of tedium, the only thing that could be stated is realism or immersion and I don't think the status quo is that immersion breaking.

At that point just tie them up first. If you haven't used up stamina you don't need to be asleep for longer than it takes to wake up, so this would be a crapshoot tactic. Regardless, the change seems meant to incentivize not being the entire round in full armor getup once you acquire it, since as mentioned its already possible to tie people up after they fall asleep and have time afterwards for a coup de grace.

With how absolute and easy to come by armor is right now, this seems honestly necessary. If it's just the chestpiece, nobody is going to die over a total of 8 seconds and 2 clicks more each time you sleep.

SaintBurgher:
Amazing change

NPC1314:
> 
> A rare amount of people wear masks and usually they only wear them to fight. And I don't find it fitting not being able to sleep in some current headgear thus why not all of them are barred from such. The issue is not even in the masks and helmets it's in the fact that nobody ever takes their armor apart to heal their wounds. This is the whole intention of the PR, to be radical. Not to add the var to the non issues 99% of the time like helmets and masks.

I dunno man, you sleep in a motorcycle helmet, hockeymask or fedora? I've slept in plate carrier quite a few times and while I dont recommend it, its less annoying than the helmet. Gas mask sleeping is nightmare fuel.
Anyways, lets test this, hopefully it works as intended.

NameChangeStonekeep:
I think hide and leather and gambeson/cloth armor can be slept in I think

SilvernKell:
it's really not necessary in a game where you have 3 needs that need to be handled at least every 20 minutes to prevent you from rping already

ThePainkiller:
> it's really not necessary in a game where you have 3 needs that need to be handled at least every 20 minutes to prevent you from rping already

# This.

We hate the fucking sleepy debuff.

Tipy1802:
I don't like it that now when you get unconscious and left alone without bleeding you can't sleep to heal if you have armour

anything86:
After playing several rounds with this, it actually feels too lenient imo, even in full medium armor. Because the check happens once your "sleepy" bar is full, you can lay down on the bed, close your eyes to start the process, then undress with eyes closed so you're simultaneously waiting to fall asleep and removing pieces of armor that'd stop it.

This is probably only mildly annoying if you're walking around everyday in several pieces of plate armor, which is the intention, so I'd say it is working perfectly.

NPC1314:
> I don't like it that now when you get unconscious and left alone without bleeding you can't sleep to heal if you have armour

This is a real issue.

NPC1314 (approved):

ThePainkiller:
`"This is a real issue"`
`>Merges it`

WHATTHEHELLISMYTEAMDOING

BadAtThisGame302:
> image image image

I think the response here was something about mongeloids, though I am not too sure.

PR #965, opened 2024-10-17
Adds a game satchel

lavillastrangiato:
## About The Pull Request

Adds a game satchel, for picking up and carrying meat, fur, hide, fat, sinew, etc. Can pick up all the related products on a single tile. You can craft one with some hide and a piece of cloth. Adds them to the hunter's loadout.

Sprites by me, La Villa Strangiato.

## Why It's Good For The Game

I feel like God's most maligned princess every time I roll hunter. I can't just be carrying around full sacks in my hands if I want to shoot some sort of creacher. "Just bring a cart, just bring a backpack!" Whose cart am I gonna cop with what small logs? What backpack am I gonna craft when I haven't hunted yet (and where is the merchant when I truly need them most?)

Also to me it feels so bizarre to be porting a cart around as a hunter. "Yeah I gotta go hunting, a form of resource gathering where you must remain stealthy and patient to catch prey. Let me just--" [sound of cart rattling on forest floor]

## Pre-Merge Checklist


Optimism333 (commented):
Seems an interesting idea. Would also be a useful item for the butcher.

code/game/objects/items/rogueitems/bags.dm
```diff
@@ -107,3 +107,77 @@
...
+		testing("yea64")
+		var/obj/item/I = pick(things)
+		STR.remove_from_storage(I, get_turf(user))
+		user.put_in_hands(I)
+
+/obj/item/storage/meatbag/ComponentInitialize()
+	. = ..()
+	var/datum/component/storage/STR = GetComponent(/datum/component/storage)
+	STR.max_combined_w_class = 100
+	STR.max_items = 30
```
NPC1314, line 139: 30 seems a bit excessive

lavillastrangiato: It *seems* like a lot, but considering the regular butcher result for a (male) gote is 11 items, it means you can go out and hunt roughly three animals before needing to return to dump out your stuff. I think it makes the hunter gameplay loop less easily interrupted by full storage capacity.

If the hunter role can also carry more, it puts them on the same level as a butcher. Let's be honest; butcher is not a super fun or engaging role, so I think making a different towner role more viable to play from a mechanical perspective helps encourage towner play more.

Optimism333 (commented):
Could maybe be a little less slot rich, let's say 25? Enough for about two animals worth. We can see how that goes, buff it a tad more if necessary.

I will also ask that butcher be given one of these.

lavillastrangiato:
This is an acceptable compromise. I'll add to the butcher's house soon, I forgot to do it on first brush.

Optimism333 (approved):
Let's get it in.

PR #987, opened 2024-10-20
Miracle Tweak

NPC1314:
## About The Pull Request

Three things affecting Miracles

* The Devotion grind for top tier spells is a little overtuned, reduced to 100 200 400 from 100 250 500. Better and also makes 
a little more sense to double every step, no more.

* The negative effects of miracle healing from the PR a while back was to the wrong spell. Fixed.
Recap is: Excommunication impotent && godless benefiting from miracles while denying they exist is a bad.
This put some mechanical weight behind excomm.

* The ported minor heal thing leant heavily into Eora being pacifist, heritage from Ratwood and heavily incentivized players to  not use their actual acolyte outfit & use a trait making you unable to defend yourself or anyone else. Like an Eoran acolyte wouldnt fight to defend their own kid or stop someone from maiming beaty? Contradicts lore pretty strongly.

Thus rtemoved the bonus for the caster weaing the pacifist enforcing flower, retains the recipient getting their bonus though since it does make sense Eora would be extra kind to gentle beings maybe, its a compromise I suppose but core issue is gone.

Also adds the dwarf combat music to the blacksmith role dwarfs, oversight they didnt get it, the most common one after all and just got generic currently.

Exommunicate part wasnt local testable so need a TM to make sure it works as intended.

## Pre-Merge Checklist


sanshoom:
Hooooly devotion grind. Speedmerge pl0x.

ThePainkiller:
Will this solve #989 ?

NPC1314:
> Will this solve #989 ?

It wont, I'm doing another not yet published one with some miracle fixes, Ill see if I can solve it there.

NPC1314:
Actually this doesnt go far enough, devotion grind is not very well designed, its to encourage a bit of prayer and delay the top spells a little I think but not sure whats won with that. What about 80 - 160 - 240 ?

Optimism333 (approved):

PR #998, opened 2024-10-21
The Dodgening

SELFHELL:
## About The Pull Request

I dislike balance coding, but people keep complaining about dodging so here we are.

This PR does two major things:

1. Makes the **Dodge-Expert** Trait check the **Armor Class** of clothing in the **Pants** slot
2. Increases the Stamina Cost of _"Trained"_ dodging to **8**, compared to the original cost of **5**

## Why It's Good For The Game

Ill keep this simple because its been said a bunch before, but at the moment dodging is objectively superior to parrying; especially an "expert" dodge. 

With Dodge Expert, you can functionally avoid every attack from multiple people for minutes at a time because of its **extremely** low stamina cost, all the while your attackers receive a further stamina penalty for DARING to attack you. Worse yet, you can still be rather well armored during this, meaning that the rare hit on the dodger will be wasted.

Increasing the Stamima Cost to more reasonable levels will help dodge return to its status as a sidegrade to parrying, rather than a direct and overwhelming upgrade for most classes.

This PR will likely require a couple rounds of TM's to iron out the exact values, as I cant do much practical testing on a localhost.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/roguetownprocs.dm
+++ b/code/modules/mob/living/roguetownprocs.dm
@@ -389,7 +389,7 @@
 				H.Knockdown(1)
 				return FALSE
 			if(H?.check_dodge_skill())
-				drained = drained - 5
+				drained = drained - 2
 //			if(H.mind)
 //				drained = drained + max((H.checkwornweight() * 10)-(mind.get_skill_level(/datum/skill/misc/athletics) * 10),0)
 //			else
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -611,7 +611,7 @@
 			spawn(used_time + 5) regenerate_icons()
 			rogue_sneaking = TRUE
 	return
-/* 
+/*
 	if(world.time < mob_timers[MT_INVISIBILITY]) // Check if the mob is affected by the invisibility spell
 		alpha = 0
 		return */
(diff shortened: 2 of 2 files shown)


sanshoom:
If this fails then I am going to blow my head off.

anything86:
Since this is targeting leg armor specifically, please take the opportunity to standardize chausses to their logical protection values

Right now iron chain chausses have steel chain protection levels (80), steel chain chausses have plate level protection (100) and plated chausses just have higher durability.

Should set iron chausses to their intended 60 protection and tag them as light, and adjust steel chain chausses to 80.

Optimism333 (commented):
Mostly agree with the feedback from anything, though I won't force you to do that in order for this to pass.

You tested this and it worked as intended?

Optimism333 (commented):
Let's give it a test, at least.

anything86:
It is also worth mentioning here, if only for future record as this was already brought up in the Discord, that the only light leg armor alternative are leather trousers, which currently have 10 melee protection as opposed to even the 30 a gambeson would get, or the 40-50 other leather armors get.
There are also no sprites for something like a studded leather equivalent for legwear; hence my suggestion of bringing at least the iron chain chausses to their intended armor value while making them light as is the case for the iron chain vest. Or just, you know, at least fix leather trousers.

SilvernKell:
yay more nerfs for dodge-users that don't deal with the issue that it's caused by racial +speed bonuses so anyone not playing a meta race is even worse off again

NPC1314 (approved):

PR #1007, opened 2024-10-24
Wound Tweaks 2: The Bloodening

anything86:
## About The Pull Request

This PR aims to fix certain issues with hardcrit and bloodloss, as well as shift the values of maximum blood and bleed in order to add harsher consequences to choosing to staying in combat while in debilitating condition, while giving people some more time to react to critical wounds so that this doesn't feel too punishing or unfair.

Also makes changes to a few other things detailed below.

Main changes include:

- Doubles the maximum blood amount of humans (as well as all associated thresholds), while doubling the bleed of normal injuries and very slightly shifting the bleed of torn arteries. This means a net increase in time before an artery wound incapacitates you, but consequently makes it harder to rouse people out of severe bloodloss. 
- Fixes the check on sleeping so that you cannot recover blood from sleeping while bleeding. Previously this was causing people to continuously recover above the critical bloodloss threshold (20 or less blood), making the speed of oxyloss be 1.5 per tick instead of the intended 4.
- Increases the total oxyloss from being in said critical condition to 5.5 per tick.
- Also adds checks on other minor sources of natural blood recovery so they don't proc while bleeding.
- Comments some unused/non-functional code associated with oxyloss at low blood levels.
- Adds bleed rates to eye gougings, tongue dissections and timpanum ruptures. Previously, these critical hits didn't do much in a fight, despite having the same weight when rolling for a critical hit. In practice it made them a detriment when aiming for those regions. You can now sew these injuries to stop the bleeding, but the intended behaviour of needing surgery or miracles to fully recover from them remains.
- Water goes back to recovering 10 blood per tick (from 4) to account for these changes.
- Corpses can now be cremated again. Checks simplified since this can only happen to corpses, only requires that the chest be at max damage and for it to remain on fire for long enough.

Other changes include

- Delimbing is now harder. The force threshold on delimb was increased significantly (10 -> 25) to match, at bare minimum, swords wielded in two hands, or the odd iron messer if on chop & strong intents. This means weapons like daggers should never be delimbing at this point.
- Decapitations in particular are less likely, specially on first strikes.
- Bad water is much less nauseating. 8u are required before puking.
- Strong poison is more nauseating, such that a poison bolt will eventually cause one to puke.
- Cranium fractures are more punishing, and will eventually cause death. Neck fractures are also given a death timer for gameplay reasons. Health potions can stabilize this.
- Vampire Lord's critical weakness is gone, but decapitation will still dust them. Furthermore, the Regenerate ability will also regenerate lost limbs. This should hopefully give them actual staying power in a fight.

## Why It's Good For The Game
Mechanically: Blood was bottoming out way too quickly from injuries, but this was ultimately meaningless due to a broken oxyloss timer. Because it was such a volatile variable (quickly going from 560, to 0, back to 560 through sleep) life-threatening bloodloss wasn't being properly simulated, making the whole system feel arcadey.

Speaking from experience, it was all too common to see a full room of incapacitated people be all brought back from critical condition despite severed arteries. These changes should remedy this, requiring the usage of health potions or immediate medical attention (bandaging, sewing, water feeding) in order to avoid death. In practice, it should mean that in the aforementioned scenario, a few people would die as more attention would need to be dedicated per person in critical condition.

The possibility of dying should also gradually make people think twice before jumping into random fights even as non-combat roles (a very common occurrence right now).

## Pre-Merge Checklist


NPC1314:
Starting to feel its time to take the W and do another PR for future stuff. Wrap it up and we merge this, overall the changes seems to have had the desired effect mostly.

NPC1314 (approved):

PR #1017, opened 2024-10-26
Undermaider & Volftrap

NPC1314:
## About The Pull Request

Undermaiden got changed to Veiled Lady in a couple places at some point, most notably the gods description, making the references in various places like spells a bit odd. Undermaiden is the classic name for her, no need to change it IMO, so just returned it to the desc.

Mantrap been annoying since forever, its a classic name but whats annoying is CONSTANTLY hearing people mistake mantraps and maneaters. It does make a lot of sense to have them named volftraps, but maybe its not grimdark enough.

## Pre-Merge Checklist

### DIFF
--- a/code/datums/gods/patrons/divine_pantheon.dm
+++ b/code/datums/gods/patrons/divine_pantheon.dm
@@ -51,7 +51,7 @@
 /datum/patron/divine/necra
 	name = "Necra"
 	domain = "Force of Death and Decay"
-	desc = "The Veiled Lady, a feared but respected Force and keeper of the dead."
+	desc = "The Undermaiden, a feared but respected Force and keeper of the dead."
 	flaws = "Unchanging, Apathetic, Strict"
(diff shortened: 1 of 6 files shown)


sanshoom:
I'm fairly certain it was changed to Veiled Lady because people keep calling necra a fucking NITEMAIDEN instead of UNDERMAIDEN

NPC1314:
> I'm fairly certain it was changed to Veiled Lady because people keep calling necra a fucking NITEMAIDEN instead of UNDERMAIDEN

We're taffed whatever way we turn arent we?

BadAtThisGame302:
Personally I don't see people confuse mantraps and maneaters and volftrap isn't doing it for me. 

The Necran change I am indifferent about, both names could be used, It doesn't matter much.

NPC1314:
Lets not change mantrap then, Undermaiden will be done in a minor fix PR.

PR #1014, opened 2024-10-25
Title + Stevedore rebranding + Keys changes

BadAtThisGame302:
## About The Pull Request

Removes the Merchant key from the Stewardry and replaces it with a Nitemaiden key.

Removes the Merchant key from the Hand and replaces it with the Captain's key.

Renames the Shepard keyring to Adept keyring. (Given it's given to the Adepts)

~~Moves Niteman to the noble positions~~ and adds a Concubine/Champion of Rockhill/Master of the Forge title.

Gives the Stevedore's a do-over, descriptions and map changed to accomodate this.

Adds 2 other unused sprites to the chance of the gold statuette.

Added an unused variant for the statue of Astrata that features more jewerly in the code. No reason for it to be left unused.

## Why It's Good For The Game

Most of the community is on the side of the Crown not getting Merchant keys so this finalizes it.

Renames to keyring is QOL ~~aswell as moving the *noble* Niteman to the noble positions.~~ Also the Concubine is a title along with Consort I originally wanted to add but withheld Consort for now, kept Concubine. Also added Champion of Rockhill for any true PATRIOT that did a heroic deed and deserves to be venerated. Likewise the Master of the Forge is a smith that's PEAKED at smithing and deserves to be YOUR boss.

Merchant is generally nowadays after it was removed from the required roles, not played much or killed because he's charging too much so that others may use his GOLDFACE freely. Pushing the Mercenaries to him didn't work either because of no mercs or no merchant or mercenaries working for the Crown more than him. So made him the Merchant's goons and pray that this works out in the end.

## Pre-Merge Checklist

- [ ] You tested this on a local server.
- [X] This code did not runtime during testing.
- [x] You documented all of your changes.

PR #1018, opened 2024-10-26
King's Rod Changes

anything86:
## About The Pull Request

Moves the king's rod functionality away from the Meister accounts and towards job titles.

## Why It's Good For The Game

The King's rod was a ported mechanic from LFWB, wherein the King could zap (and stun) anyone wearing a Cerberii helmet. It was intended as a way for the Lord to enforce his will onto the Garrison and keep them in line. As ported to this codebase, it applied to anybody with a Meister account, which meant most of the town. It slowly cultivated a meta culture of Kings sometimes creating laws at roundstart demanding people create Meister accounts in order to not be persecuted by the Garrison, mostly because the rod's hardstun can end combats immediately and without recourse, and Meister accounts could not be deleted by any means.

Most egregiously, this applied even during peasant rebellions, court coups and church and crown conflicts. The rod had the potential to end large group combats in an extremely uninteractable way.

With these changes, the King retains his intended control over not only his Garrison, but also the court and servants, and is still be able to shock people if he elects to change their job to any of those in the list through the throne function (one of which is Prisoner, if so needed).

## Pre-Merge Checklist


anything86:
This can be testmerged whenever, but I'll want to move the list to jobs.dm eventually before merging. Right now #1014 is open so that would could create conflicts.

code/_globalvars/lists/objects.dm
```diff
@@ -40,3 +40,24 @@ GLOBAL_LIST_EMPTY(ai_status_displays)
 
 GLOBAL_LIST_EMPTY(mob_spawners) 		    // All mob_spawn objects
 GLOBAL_LIST_EMPTY(alert_consoles)			// Station alert consoles, /obj/machinery/computer/station_alert
+
+GLOBAL_LIST_INIT(rod_jobs, list(
```
BadAtThisGame302, line 44: You need to add all the noble_positions and the Mayor.

anything86: Added. Not a fan of the Niteman being included even if he is a noble. Priest also has a manor key but is not included. Would appreciate further feedback on both of these.

BadAtThisGame302: I think Priest should realistically be on it. He is a local and rules the Church, the King does need something over him maybe? And any thoughts on the Merchant being on it?

anything86: Merchant could be included since realistically he's not a fighter and if he were to oppose the crown he would do it with mercenaries. Priest is not a fighter either.. It would probably be fine so long as the rest of the church isn't included. I'll add both.

NPC1314: It seems potentially bad to keep priest and merchant roddable but not the inquisitor or innkeeper etc.
Neither role being very popular, and the game would be more interesting if they had a little mroe chance to be opponents to the King at times.

Just keeping it to people who spawn in keep seems cleaner, neater in concept IMO, if it needs changing at all.

NPC1314:
Lets see how it works in practice then

Dagonlives:
If the goal is to nerf the rod and make it more applicable to put the king's people in line, let it be usable on the keep roles only.  If we are going by lifeweb comparison, only 'knights' and 'garrison' would be targetable with the rod.  I'm fine with seeing court wizard and other court roles added to that. 

Church, merchant or other non court roles shouldn't be targetable.

NPC1314:
> If the goal is to nerf the rod and make it more applicable to put the king's people in line, let it be usable on the keep roles only. If we are going by lifeweb comparison, only 'knights' and 'garrison' would be targetable with the rod. I'm fine with seeing court wizard and other court roles added to that.
> 
> Church, merchant or other non court roles shouldn't be targetable.

That is essentially how it is implemented now is my understanding so basically just need confirmation it works as intended for a fullmerge

NPC1314 (approved):

PR #1020, opened 2024-10-27
Lonely Priest Seeking Choir

NPC1314:
## About The Pull Request

Not entirely happy with where Temple is at, Priest gets bullied a little too easily. 

* Priest spawns with their staff inhand
* Aasimar and Dwarf priests gets to try the new Priest helmet I made. It got antimagic included so they don't need the null magic ring any more, Astrata got them covered as long as they wear their hat like theyre meant to.
* New Priest hat lets you toggle the mask to show your name and descriptors without taking off your insignia. Also get a tiny bit of armor. The magic defense makes you immune from other Astratans using their fire spell on you
* Priest dying triggers a single Bad Omen (doesnt repeat like the Ruler one, needs testing a few rounds just to make sure its not buggy, but worked fine in local)
* The Astra fire spell slightly bufffed by having 2 sec longer burntime, but slightly lower range and longer cooldown. Until fire damage is fixed this has to do.

## Pre-Merge Checklist


code/modules/jobs/job_types/roguetown/church/priest.dm
```diff
@@ -39,22 +39,30 @@
 	shoes = /obj/item/clothing/shoes/roguetown/shortboots
 	beltl = /obj/item/keyring/priest
 	belt = /obj/item/storage/belt/rogue/leather/rope
-	beltr = /obj/item/storage/belt/rogue/pouch/coins/rich
-	id = /obj/item/clothing/ring/active/nomag
+	beltr = /obj/item/rogueweapon/mace/church
```
BadAtThisGame302, line 42: Two things, first this spawns on the map and shouldn't be duplicated. 

Second, the Priest DOES have a staff and this is the path, feel free to check it out. /obj/item/rogueweapon/woodstaff/aries

NPC1314: ugh,I'll just put it in their hand instead then, no point having their staff laying about

BadAtThisGame302 (changes requested):
Don't keep the bell ringer, it spawns next to the bell on every map, same as the staff. Please, this is duping.

NPC1314:
> Don't keep the bell ringer, it spawns next to the bell on every map, same as the staff. Please, this is duping.

relax, just forgot to delet. The mapped staff will go as soon as this goes into testing too, dont worry be happy.

Optimism333 (commented):
image

Slight mistake, you added a second line of polearms skill here.

Optimism333 (approved):

PR #1023, opened 2024-10-28
Revert "Treasury payment/fines fix (hopefully)"

NPC1314:
Reverts Darkrp-community/OpenKeep#962

Since it breaks logging and no one will fix logging its not worth the cost after all.


NPC1314:
Works, good for merge.

Optimism333 (approved):
Hopefully in the future we can add fining without breaking something else.

PR #1024, opened 2024-10-28
Dunmanor Mapgen updates & Ship

NPC1314:
## About The Pull Request

So Dunmanor got some runtimes with z-level stuff, thats why the mapgens on main doesn't work right.
What seems to work is putting sunlight on z-level 3 instead of 4, so lets see if it works on live too so we can skip manual lights forever.

* Adjusts sunlight gen to z-level 3. Needs TM to see if it works, its last chance so if not its time to go back to manual lights

* Bandit camp got a new quick route back home from the main map north side. So bandits doesnt have to always go via the bottleneck forest entry to get home

* Bandit river dried up a bit, now you can actually jump across without putting out your torch in some places

* Coastline adjustments to make it possible to climb back up if you fall down and to get sunlight down there where it should be.

* Couple varedited travel tiles replaced with properly coded ones on centcomm, shouldnt have any visible effect.

* The late arrival ship was too big, to empty. to clean, it lacked some charm so I edited it and added some sounds etc.
Its more interesting to spawn here  now IMO. see video.

* Adds some underworld maze stuff, rudimentary, but couple coins at least

* Bandit lair and Vampire manor got their areas adjusted to remove maneaters and such randomly spawning there.

https://github.com/user-attachments/assets/e3ca9116-4fbd-42cf-95af-623cb3f7a930

## Pre-Merge Checklist

### DIFF
--- a/_maps/map_files/generic/CentCom.dmm
+++ b/_maps/map_files/generic/CentCom.dmm
@@ -666,9 +666,7 @@
 /area/rogue/outdoors/woods_safe)
 "bE" = (
 /obj/machinery/light/rogue/smelter/great,
-/turf/open/floor/rogue/ruinedwood{
-	icon_state = "weird1"
-	},
+/turf/open/floor/rogue/metal/barograte,
 /area/rogue/indoors/vampire_manor)
 "bF" = (
 /obj/structure/fluff/walldeco/maidendrape{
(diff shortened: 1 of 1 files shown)


NPC1314:
Ahh finally, this solved the mapgen thing. For future reference, the runtimes in DunManor z-level setup makes sunlight not shine down as it should from z-level 4, so this uses level 3 instead.

Also fixed couple minor issues like not being able to cross bandit river dry, reducing amount of kneestingers, maneaters, volftraps in the forest area, getting the right trolls to ambush in the right biome.

Optimism333 (approved):

PR #1038, opened 2024-10-30
Fixing hats

NPC1314:
## About The Pull Request

We hitting rock bottom here with PR puns.

Anyways, more Temple fixes/upgrades/leaning into making the faiths more different is the continued theme

* Priest latejoins Astratans now, not some Godless rando.
* Added a proper path for the inhumen camp, its a flooded cave in the north
* Added inhumen camp traits to wood elves and orcs and goblins so all the scum can congreate in their shitty camp
* Alt hat for necrans mapped into crypt, the staff priest spawns with no longer duped
* Priest/Acolytes gets mace 2/1 respectively so they actually can use the ringer instead of their staff if they want to.
Because the staff looks pretty terrible on your back, the priest one is legit dreadful, zealots needs a drippier alternative even if worse skill.
* Eoran curse reworked, makes you drunk as hell and poisoned but doesnt heal you and give you moondust any more because it was too weird before. Now its a debuff and a stupid curse to haunt the victim for a good while. Truly you get the effect of drinking tons now. 
* Eora acolyte being able to marry people - its being tested, hard to test locally so can't guarantee it works but got a couple version to try. Possibly will make priests unable to do marriages, so need to do a supervised TM to make sure.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/special/crusader.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/special/crusader.dm
@@ -151,6 +151,8 @@
 /obj/item/clothing/head/roguetown/helmet/heavy/crusader/t
 	desc = "A silver gilded bucket helm, inscriptions in old Psydonic are found embezzeled on every inch of silver. Grenzelhoft specializes in these helmets, the Totod order has been purchasing them en-masse."
 	icon_state = "crusader_helmt2"
+	icon = 'icons/roguetown/clothing/special/crusader.dmi'
+	mob_overlay_icon = 'icons/roguetown/clothing/special/onmob/crusader.dmi'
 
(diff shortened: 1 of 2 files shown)


NPC1314:
> Please don't implement the Deathface, it looks so fucking bad. The icon is literally a near 1:1 of the Necran hood.

Thats just the placeholder sprite, the real one is in the dye bin update.
Either way its not added because it looks better, its added for mechanical reasons since a alternative headwear that doesnt conceal the face should be an option, if you want to try to make something cooler go right ahead, that is IMO a lot more useful than more helmets.

Other stuff thats needed is a better arming cap icon, could try your hand at that too, we only got the placeholder atm.

Optimism333 (approved):

PR #1043, opened 2024-11-01
Vampire Fixes and Reverts 

BadAtThisGame302:
## About The Pull Request

Firstly and largely, reverts drinking 5 blood a minute to make turning someone into a spawn actual fucking torture.

Also, another big one, reverts this entire vitae pool thing, I never liked it, I said so on the discord, it passed, we played the Vampire Court round and it was horrific. If I, the Vampire King, that has every single person in town under my whim and can drink them all for their vitae pool.... I still could not reach max level and tear out the sun, during the event. So imagine an actual VL doing that, yeah, impossible, right?

Another big one, regen is no longer an aheal and no longer breaks your night vision eyes, thank me later.

## Why It's Good For The Game

First one is complete torture to play with and I will continue to make PRs about it, frankly. I don't care if this gets closed, I won't stop.

Second, it's unreal how even the King of Rockhill being a VL could not give me enough vitae to first of all ascend beyond level 3 (me and the Queen were both level 2, so technically me alone would be level 3) and not to mention the fact that once your vitae runs out, that's it, it never regenerates, so good fucking luck looking for people for a crumb of vitae.

And third one, it's an obvious bug, bundled in as vampire changes.

## Pre-Merge Checklist


ThePainkiller:
Denied. Vitae pool main purpose was precisely to avoid people grabbing just ONE human and keeping them as their main source of food by spamming feeding them water until they fully ascended by just capturing one unfortunate soul, no matter if they were AFK or disconnected from the game or whatever, as long as they kept that person fed and with high blood levels they could ascend without having to take more than one prisoner or convert more than one towner. It was gamey and bad.

Removing things "because you do not like them", specially things that are meant to balance the game to prevent powergaming, is not a valid argument. You could've at least done something to remedy the supposed "issues" this caused (Of which I can only think VLord on lowpop being an absolute nightmare since you can get like five people at most, or go loud from the start), like upping the vitae pool on mobs or whatever.

Players have also complained about vampire heal not fully regenerating limbs, which the coffin aheal should do, but I agree the combat spell/skill/free aheal they get should be nerfed to NOT regenerate lost limbs or organs in the middle of combat yeah. That might be the **only** thing I agree with on this PR.

**No nuclear options.** Not even maintainers do those unless it's extant mechanics we can totally do without. Either propose an actual fix, or talk it out with the noble focus group to get input on how vampire lord plays out on lowpop and the underlying issues of a vitae pool making lowpop vampire hard, or find an alternative so that you can just outright kill someone another way to revive them as a thrall, because the changes were only applied for vitae, not blood sucking out of a person. You may have to look at the recent blood pool changes someone else did for that ammend.

NPC1314:
But at the end of the day we remain with a lot of issues in VL, the fix to regen NV and aasimar and tiefs breaking I did was too troublesome, disabling the ability to customize hair etc on your VL char, kinda bad.

anything86:
I take (partial?) responsibility for breaking night vision on regeneration, though the issue with Aasimar/Tieflings was older than that. My suggestion if this is trimmed to just a few bugfixes would be adding only the limb regeneration parts of the full aheal proc to the vampire regeneration proc, and removing the admin_revive = TRUE flag which currently serves that purpose (essentially, tack the regenerate_limbs() proc onto it).

sanshoom:
If you cannot fully ascend as the VL, even when you have a horde of subjects that are literally giving blood to you.. why even enable VL? 

It sounds like an utterly asinine state to even play Vampire Lord and his goons, when compared to Zizo Cultist (because when cultist fully ascends, they are essentially unkillable).

Dagonlives:
I don't think this P.R is a good idea.  We had issues with the vampire lord before and it's 'capture a single person then show up ascended 20 minutes later' gameplay loop. 

Others have expressed my views her more eloquently.  Perhaps the vampire lord needs tweaks and improvements to it's game play, but this is not how to achieve it.

NPC1314:
Upping vitae pools modeslty (say 20% ) and fixing regen seems like the best parts of this idea.
(see anything 86 comment on the regen)

A full revert is not going to happen, we can't go back to turbobroken VLs.

PR #1046, opened 2024-11-02
Providing More Alchemy Supplies + Roguetown Soil Alignment

lavillastrangiato:
## About The Pull Request

- Adds some extra basic ingredients, such as the rune dusts and some wheat chaff, to the feldsher's, court mage's, and archivist's alchemy labs.
- Also arranges the soil lines of the farm on Roguetown so that they can line up with the plow and be, well, plowed.

## Why It's Good For The Game

To make a long story short, not a lot of people are using the alchemy system. This makes me sad. I feel like if there were more alchemy tools available at roundstart, and it was more accessible, people would use the alchemy system more, and perhaps have more fun with it. I think we should allow people to mess around with systems more, and have more fun. Sure, exploration and gathering is part of a gameplay "loop", but if the entire round is spent gathering, then that doesn't leave time for the actual crafting bit.

## Pre-Merge Checklist


NPC1314:
Sounds reasonable enough. lets see if it can be added without mucking up the book TM

NPC1314 (approved):

PR #1052, opened 2024-11-03
Comments melee embeds and nets/bola acquisition

anything86:
## About The Pull Request

Sets generic melee weapon embed chance to 0, removes the 75% embed chance on artery critical hits, comments out the recipe for nets as well as option to buy them at the goldface.

## Why It's Good For The Game

Melee embeds are currently incredibly buggy and/or poorly implemented. They have distinct behaviors depending on whether it is a carbon mob or simple mob.

On carbon mobs, they won't remove the weapon from your hand. This means the weapon is now both embedded in the victim causing extra bleed, as well as in your hand and available to continue attacking with. If removed from the victim, this will teleport the weapon out of your hand no matter how far you are, potentially glitching out entire inventory slots (if on your back, you now need to ask someone to remove it to fix it. If on your belt, removing the belt is usually enough to fix it). 

On simple mobs, the weapon will vanish from your hand entirely, initiate a grab (which gets broken if, say, in the heat of combat you move away) and add two intents tied to the grab. This is a problem because, as previously mentioned, should you move away which can happen in an instant, your weapon is now gone and retrievable only by a deeply unintuitive process of examining the mob, clicking "inspect wound", and then clicking the weapon as it shows on the chat box.

Given these current issues, it seems more prudent to remove the possibility of melee weapons embedding until they're ironed out.

--

Nets (henceforth referred to as bolas) are poorly implemented, uninteractive and similarly prone to glitching. 

For starters, due to some unholy fusion of reutilized handcuff/legcuff code and its associated massive action cooldown when you attempt to resist out of them, being hit by a bola will shut down your ability to click for several seconds after resisting out of one. This means that even if you reacted fast enough to immediately struggle out of a bola, your assailant can now wail on you as you find yourself confused (through your monitor) and unable to do anything (in the game). This also means that should you find unfortunate enough to be face to face with an absolute human specimen intent on throwing MORE than one bola at you, you CANNOT resist out of the second bola that hits you for a similar amount of time.

Secondly, getting hit by a net knocks your inhand weapon to the ground. Add to that the point above, and you can't even retrieve your weapon after struggling out of the net. The person who just tossed it at you might as well run up to it, grab your weapon, and you're now disarmed.

Third, it has a massive multiplicative slowdown of 3 on top of the associated FIVE (!!!) speed (and two endurance!) debuff. Why? Fuck you buddy, you don't get to ask questions here.

Fourthly, ANY motherfucker can craft this with a rope and stones, or buy THREE for fifteen mammon. ROPE AND STONES! THEY'RE EVERYWHERE! Be glad the average player is not intent on making use of this sadistic contraption.

Fifthly, there is even a piece of code attached to the fucking hitby proc specifically designed to check IF you were hit by a bola, so that you cannot catch it by readying throw. Counterplay is for nerds! There's even a comment right next to it calling you a dumbass!

With all that said, nets deserve to be banished to the shadow realm. Or to the very specific female anthraxi spawn. Nobody plays that anyway and I forgot to remove that line.

## Pre-Merge Checklist


anything86:
Further local testing revealed that bolas will knock your weapon out of your hand still. I was also able to replicate normal behaviour on localtest, and confirm that it stops you from clicking at all for a long time after slipping out of one. Sadly I cannot localtest to check if the old glitch that permadebuffs you if you catch a net is still around.

KreepyKommy:
It seems medical tools cannot embed as well due to this PR. This means speculums and forceps can't be used at all making almost all surgeries impossible.

NPC1314:
Unmerrged until fixed due to the issue repoorted, people need to be able to surgeryize

anything86:
Confirmed surgery is working. Scrubbed it off the last place it can still spawn in as well.

NPC1314 (approved):

PR #1057, opened 2024-11-04
Addresses some issues in delimb breakpoints, makes "easy dismemberable" mobs slightly more vulnerable to dismembering.

1N7Slayer:
## About The Pull Request

This should sort out perfectly sharpened iron cutting weapons that are at least two sizes larger than daggers that could not remove limbs prior. Daggers still won't be able to do it, except possibly for the profane assassin dagger against a hunted target - however, it is unique as a weapon and limited in its scope, if it actually manages to delimb somebody that's basically the equivalent of picking the Bloody Mess trait/perk in Fallout. Skill issue if you take the hunted flaw and this happens to you, tbh.

## Why It's Good For The Game

The threshold of 25 compared to 20 was just too severe. Iron swords were not able to take the heads off of goblins even after 20 or so strikes, whom were already critically weak to delimbing in the first place! This should address that problem while also not letting _most_ dagger users take advantage. As the steel dagger is 14 damage, you'd need at least 15ST (14*0.5=7, 14+7=21) to make that even remotely threaten decapitation or other forms of dismemberment.

Against non-goblin targets this should bring arm-length cutting weapons back into threatening viability of potentially being able to take off limbs again without STR intent instead of heavy cutters (longswords, greatswords, battleaxes, poleaxes) being the only ones to reliably be able to do this.

## Pre-Merge Checklist


NPC1314:
Lets try this, it does seem 25 was too hard a hit, but maybe the best value lies at 22 or something like I suggested a while back.
anyways, getting a TM to see how it plays out, current situation not good.

anything86:
Note that strong intent multiplies the end force by a further 1.1 for the threshold. I think 20 specifically is bad because it allows for one-handed swords like the rapier to delimb at 10 strenght without even being on strong intent. It is still incredibly consistent to just slash a limb a few times and have it pop off if its above the threshold.. I think the formula is just bad, honestly.

If its for NPCs then it's fine, so maybe an if statement that checks if they have the easy dismember/critical weakness trait and then sets the threshold lower?

anything86:
Alternatively, could try to make delimb respect crit protection.

1N7Slayer:
> Another issue in general is just how much people seem to need to rely on delimbs for goblins. You can even disembowel one and it'll keep going, just very frustrating.

the likely reasoning for this is that pain is chance based every tick instead of threshold based - and you can't collapse in pain if laying down, since you're already collapsed. like there should be a point where you're suffering so much pain that you're effectively paralyzed and need healing to have any chance of functioning further.

NPC1314:
With how simple it is to tweak this number and how terrible delimbing is working in localtesting lets just move along.
The struggle to make combat less buggy goes on.

NPC1314 (approved):

NPC1314 (approved):
its not this PR so its fine

PR #1058, opened 2024-11-04
Armor system

NPC1314:
## About The Pull Request

Current system lacking
*Basic concept is the three armor classes also got a span of defense they cover so medium will on average always have higher values than lights etc.
*Values generally somewhat lower so more attacks can pierce partially without having to game so hard.
* Crit protection more valuable
* Goes hand in hand with the tweaking of wounds happening and the overview of weapons to make them sensible (make useless ones merely bad, make the nobrainer ones slightly less OP)
* no longer same armor have different AC, so maille is medium only no matter iron or steel

Will obviously need tweaking and adjustments and testing, but the concept is solid. Its vastly more important things make sense for a player who looks at armor mostly of metal to know "ok this is roughly that sort of tough and its not good for dodging" without having to codedive, instead of the materials following a strict system. 

## Pre-Merge Checklist

### DIFF
--- a/code/modules/clothing/rogueclothes/neck.dm
+++ b/code/modules/clothing/rogueclothes/neck.dm
@@ -222,7 +222,7 @@
 	max_integrity = INTEGRITY_STRONG
 
 /obj/item/clothing/neck/roguetown/gorget/copper
-	name = "copper gorget"
+	name = "neck protector"
 	icon_state = "copperneck"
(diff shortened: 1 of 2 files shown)


anything86:
So far I assume this is only affecting chest armors and helmets? Any plans to extend it to gloves, legwear, boots, etc?

NPC1314:
> So far I assume this is only affecting chest armors and helmets? Any plans to extend it to gloves, legwear, boots, etc?

Yeah, anything with an armor value really.

Optimism333 (approved):

PR #1069, opened 2024-11-06
Dodge/Parry Experimental Tweaking

NPC1314:
## About The Pull Request

Testing how to make weapon balance a bitbetter.
currently it only matters when you got a higher speed than your opponent, in wich case it impacts dodge modifiers and it multiplies with the speed difference. Maybe "realistic" but its partly to blame for very high dodoge values and even at same speed, a slow clumsy weapon shoukd be easier to dodge.

So tried to adjust the fomrula for that effect. Did I succeed? I dunno.

EDIT
Seems to be working ok, added two more things besides defines to easier tweak the dodge/parry formulas 
Goal is to get more outcomes that doesnt hit the 95% ceiling, its almost the norm still.

1. Weapons skill weight on parry chance reduced by 5%. Should have marginal effect, might need some plateus but it ever so slightly should nudge towards the weapon defense mattering a bit more.

2. The dodge expert bonus to dodging reduced to 10% from 20% per speed point. While the formulas are somewhat opaque hitting 100+ percent is too easy generally speaking, this might be too much though, value might need more tweaking, but maybe not. Lets see how it plays out.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/mob/living/roguetownprocs.dm
+++ b/code/modules/mob/living/roguetownprocs.dm
@@ -1,7 +1,7 @@
 // PARRY AND DODGE FORMULA DEFINES : for easy adjustments. These multipliers are integrated into the formulas, can be set to different values to adjust desired outcome likelihoods
-#define PARRY_SKILL_WEIGHT 18				// was 20
+#define PARRY_SKILL_WEIGHT 20
 #define WEAPON_DEFENSE_WEIGHT 10
-#define SPEED_WEIGHT_FOR_DODGE_EXPERT 11	// was 12
+#define SPEED_WEIGHT_FOR_DODGE_EXPERT 12
 #define SPEED_WEIGHT_FOR_DODGE 10
 
 


code/modules/mob/living/roguetownprocs.dm
```diff
@@ -378,10 +378,12 @@
 		if(U)
 			prob2defend = prob2defend - (U.STASPD * 10)
 		if(I)
+			prob2defend = prob2defend - ( I.wbalance * 10)	// so slow weapon gives +10 % chance to dodge and fast the opposite. So it doesnt get multiplied too by the dodge bonus of being faster.
```
ThePainkiller, line 381: Worrided about this one doing two negatives a positive when the weapon is heavy, as the final formula would be:

`prob2defend - (-10)`

And you know how BYOND loves to fuck with our perception of math. I say we give this a TM to test and see. Have you tried giving weapons to bums and defending with them on testing?

NPC1314: It could be it doesnt add up, I haven't tested a blessed thing for sure.
- - shoule become plus which is what we want but only BYOND knows for sure

anything86:
Assassins already had only 56% dodge chance against roles such as Captain, or 68% against ruler Queen, assuming same race. I think this will affect these percentages at the top end, while still doing nothing to uncap dodge rate against non-combatants and more average fighters. All in all I just don't know what this is supposed to fix.

NPC1314:
> Assassins already had only 56% dodge chance against roles such as Captain, or against ruler Queen, assuming same race. I think this will affect these percentages at the top end, while still doing nothing to uncap dodge rate against non-combatants and more average fighters. All in all I just don't know what this is supposed to fix. Shifting the dodge expert weight altogether will represent a flat -12 to -14% dodge chance across every dodge chance.

The reason is stated in the PR, the weight of dodge expert being tested as +10% instead of +20% per SPD does not do what you say, and it doesn't  even matter. 
Its a test thats concluded, no point going on with no data.

PR #1073, opened 2024-11-07
Revert "Skullcrush instakill hotfix"

NPC1314:
Reverts Darkrp-community/OpenKeep#791

So how about that

### DIFF
--- a/code/datums/wounds/fractures.dm
+++ b/code/datums/wounds/fractures.dm
@@ -83,7 +83,7 @@
 		if(iscarbon(affected))
 			var/mob/living/carbon/carbon_affected = affected
 			carbon_affected.update_disabled_bodyparts()
-	if(mortal && HAS_TRAIT(affected, TRAIT_CRITICAL_WEAKNESS))
+	if(mortal || HAS_TRAIT(affected, TRAIT_CRITICAL_WEAKNESS))
 		affected.death()
 
 /datum/wound/fracture/head/on_mob_loss(mob/living/affected)


BadAtThisGame302:
Why is this revert good?

1N7Slayer:
> Why is this revert good?

We never got the bandaid fix Nick was talking about to make hits after the skullcrush a death blow is perhaps the main reason

NPC1314:
Theres been a ton of complaints for months about blunt being unable to kill NPCs, this is the obvious culprit given what it touches and the date.
If goblins etc dies now as they should, then this revert gets merged.
If it does not help, it not getting merged. 

Simple as.

1N7Slayer:
> Theres been a ton of complaints for months about blunt being unable to kill NPCs, this is the obvious culprit given what it touches and the date. If goblins etc dies now as they should, then this revert gets merged. If it does not help, it not getting merged.
> 
> Simple as.

To my understanding stabbing goblins in the heart was the way to go, but indeed other weapon types were not able to kill them.

NPC1314:
> > Theres been a ton of complaints for months about blunt being unable to kill NPCs, this is the obvious culprit given what it touches and the date. If goblins etc dies now as they should, then this revert gets merged. If it does not help, it not getting merged.
> > Simple as.
> 
> To my understanding stabbing goblins in the heart was the way to go, but indeed other weapon types were not able to kill them.

I noticed the same with zombies, just mauling their head endlessly with a huge club did nothing, its lame and I suspect this PR is to blame. But it should be obvious what the truth is with some more testing

1N7Slayer:
> > > Theres been a ton of complaints for months about blunt being unable to kill NPCs, this is the obvious culprit given what it touches and the date. If goblins etc dies now as they should, then this revert gets merged. If it does not help, it not getting merged.
> > > Simple as.
> > 
> > 
> > To my understanding stabbing goblins in the heart was the way to go, but indeed other weapon types were not able to kill them.
> 
> I noticed the same with zombies, just mauling their head endlessly with a huge club did nothing, its lame and I suspect this PR is to blame. But it should be obvious what the truth is with some more testing

Perhaps looking at what causes death() with heartshots against critically weak mobs should be transplanted onto the top-tier-most critical injuries associated with blunt weapons would be a better change instead? I feel like all non-carbon mobs should be critically weak on account of them operating on pointmed, as carbon mobs can collapse from pain or become incapacitated by limb damage, whereas NPC simple mobs can continue attacking until dead.

anything86:
I was crushing orbital bones fine during the halloween skeleton siege. It caused skeletons to drop to the ground and stop moving. Perhaps some localtesting should be done to make absolutely sure that this is not working in some way?

NPC1314:
> I was crushing orbital bones fine during the halloween skeleton siege. It caused skeletons to drop to the ground and stop moving. Perhaps some localtesting should be done to make absolutely sure that this is not working in some way? There is not even an issue open right now detailing the problem.

https://github.com/Darkrp-community/OpenKeep/issues/814 is the one issue open.
Theres rarely issues opened, like 95% of bugs never get there. Its mostly fumbling in the dark with no input.
Theres almost no documentation whether 791 actually did a damn thing or not, as with most things.
I urge anyone who wants to test out things and type up a update on PRs or issues of this type to do so, it helps immensly.

NPC1314:
Probably not needed then? Possibly? Who knows anything, what is even time.

PR #1081, opened 2024-11-08
Vampire: the Stakening

anything86:
## About The Pull Request

A couple of changes, bugfixes and tweaks to vampires and associated items/roles. Overall intended as a buff to a deprecated antagonist.

- Adjusts vampire regeneration to avoid bugs with night vision. Should now regenerate limbs, a necessity so that vampires are not completely disabled after a wound that takes realistically only 3~ slashes to inflict.
- A punctured heart will now cost a Vampire Lord 2000 from its vitae pool. Added now that vampires are not instantly dusted from heart crits in order to give them some utility. Spawns are not subject since their vitae pool is already comparatively tiny.
- Humans go from 5000 vitae to 8000 vitae. Previously, 12 fully drained humans was the bare minimum to ascend, not accounting for any blood expenditure. This should hopefully make ascension not as unattainable of a goal. Subject to change.
- Bite wounds now have a tiny amount of bleed associated, just enough for any vampire's bite to cause the location to be drinkable. You'll no longer need to stab your victim first like a fangless rat.
- Drinking blood from a human should now cause more bloodloss, regardless of whether their vitae has been depleted or not. Intended mostly as QoL for turning those who are willing.
- The Witch Hunter adventurer class gains a silver whip in addition to their silver rapier, to go with the theme.
- Whip's crack intent goes back to 20 armor penetration.
- Whip's lash intent now has unique associated wounds, including a critical. Whip lashes cause more pain and bleed than a slash, but in exchange the associated intent has very little AP, must be charged and cannot sever arteries. The critical is permanent scar tissue (removable only by miracles), perfect for punishments. Careful not to whip too hard; large whip wounds can bleed someone out very quickly.
- Silver's effect on vampires has been changed. Hardstuns are gone, as they were simply too harsh, replaced with knockdowns. Burn damage on spawns slightly toned down, weapon-dependent.
- Churn Undead and Heal should no longer hardstun when cast on undead. They retain a knockdown effect. Heal's burn damage lowered from 100 to 50, still sets on fire. Lesser Heal burn lowered from 30 to 25 to keep in line. Anastasis left untouched.
- Wooden stakes now have dagger intents attached. Includes the before-commented dagger pick intent, with hefty AP. Should be balanced out by the tiny base damage (best used on Aimed or Strong intent), still much, much worse than a dagger against proper armor. You can now drive a stake through a vampire's heart!

TO DO:
- Other vampire changes? Feedback is welcome.

## Why It's Good For The Game
Vampire Lord is by far the least opted-in antagonist, to the point that even if you have it toggled on and the gamemode is rolled, you'll likely be without any vampire spawns to assist as people simply don't turn it on. The reasons, summed up, are that it is simply not a rewarding and satisfying antagonist to play. These changes should help bring Vampire Lord up to speed after the associated nerfs months ago, and encourage more players to try it.

## Pre-Merge Checklist


sanshoom:
Would it be possible to give the Witch Hunter a steel cuirass? He is essentially defenseless since you can’t parry with a whip at all.

Another thing, wasn’t the silver rapier made specifically for witch hunter? It feels a bit odd to see it removed.

anything86:
Fair point. Could try giving him both the whip and rapier, but there has to be a slot for the pouch given the lack of a satchel.

sanshoom:
> Fair point. Could try giving him both the whip and rapier, but there has to be a slot for the pouch given the lack of a satchel.



> Fair point. Could try giving him both the whip and rapier, but there has to be a slot for the pouch given the lack of a satchel.

Could try just.. giving them a satchel to put their coinpurse in? I think they’re one of the only ‘rare adventurers’ that lacks a bag, but them having two weapons might be pushing it.

code/datums/wounds/special.dm
```diff
@@ -274,3 +274,34 @@
...
 		)
+
+/datum/wound/scarring
+	name = "permanent scarring"
+	check_name = "<span class='userdanger'><B>SCARRED</B></span>"
+	severity = WOUND_SEVERITY_SEVERE
+	crit_message = list(
+		"The whiplash cuts deep!", 
+		"The tissue is irreversibly rended!", 
+		"The %BODYPART is thoroughly disfigured!",
```
ThePainkiller, line 285: Are you sure `%BODYPART` will work here without being between brackets?

anything86: It's copied from fracture, worked fine in testing.

code/modules/mob/living/carbon/carbon_defines.dm
```diff
@@ -73,4 +73,4 @@
 	var/domhand = 0
 	var/tiredness = 0
 	// How much total vitae a vampire can absorb from this mob. Once expended, you can't gain more from them.
-	var/vitae_pool = 5000
+	var/vitae_pool = 8000
```
ThePainkiller, line 76: I never intended the vitae pool to be this big, but we'll see how quickly and easy vampire lords can ascend with this much vitae. I just don't want them to gain just two people and suck them completely out of vitae to ascend.

sanshoom:
Alright I just noticed this, but WITCH HUNTER for the longest time has had the same chance probability to roll as FULLPLATE ADVENTURERS. I personally don't think this should really be the case, so maybe it should be upped to.. 25 or 30? Same as Qatil or Vaquero, both low-armored rare classes.

NPC1314:
Any final edits?

NPC1314 (approved):

PR #1108, opened 2024-11-13
blade balance

cornercube:
## About The Pull Request
knives were all the same again, this differentiates them a bit

hunting knife deals 11 damage, 1 more than regular knives, and has STRONG integrity
daggers deal 12 damage, 1 more than the hunting knife, but only has STANDARD integrity
steel dagger deals 13 damage, 1 more than iron dagger
so daggers are pointier and fragile, the hunting knife is thicker so is stronger but less cutting
steel alt dagger has a value of 25, slightly higher than the usual, but otherwise performs the same

steel short sword is available again, sprite slightly shortened to make it look shorter than a regular sword
short swords slightly buffed too
steel short sword deals 16 on slash and 20 on stab
iron short sword deals 15 on slash and 18 on stab (previously 13 and 18)
keep in mind, it cannot be wielded for damage bonus

## Why It's Good For The Game
no one ever used short swords, this gives it a niche as a stab machine
the steel short sword was removed, but now its back

make each knife slightly different

## Pre-Merge Checklist


NPC1314:
Will have to wait for the tool PR test to finish due to icon conflict otherwise.
Looks acceptable, only real issue is you forgot to reassign so iron shortsword melts into iron and steel into steel.

PR #1110, opened 2024-11-13
Some fixes

NPC1314:
## About The Pull Request

*shalal had color issues due to not using the right palette, adjusted
* messer and other in scimitar class swords had couple bugs, fixed
* Veiled Lady restored to Undermaiden so the various descriptions arent so confusing and because there was no need to change it in the first place. Only a few instances existed.
* A couple skelly loadouts added and some cleaning etc

## Pre-Merge Checklist

### DIFF
--- a/code/datums/gods/patrons/divine_pantheon.dm
+++ b/code/datums/gods/patrons/divine_pantheon.dm
@@ -51,7 +51,7 @@
 /datum/patron/divine/necra
 	name = "Necra"
 	domain = "Force of Death and Decay"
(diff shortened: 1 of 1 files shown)


lavillastrangiato:
i don't think the veiled lady changes are really necessary, gods can and should have multiple titles

NPC1314:
I guess I can add it as an extra next to Veiled in the descriptor, because not having Undermaiden in the description when its referred to in  various places like spells is annoying. It needs to be included.

PR #1132, opened 2024-11-21
Armor/Weapon Post-Rework Second Pass

anything86:
## About The Pull Request

I started writing this PR as just some changes to make gloves and masks use the new defines, then it snowballed into this after localtesting revealed some issues with the current system.

In summary, these changes are meant to iron out the new armor system, as well as the weapons that interact with it. The changes are as follows:

### Armor:
- All relevant armor pieces should now be using the latest defines. Masks, gauntlets and boots are no longer outliers.
- Integrity defines were changed such that the tiering goes 200 -> 250 -> 300 -> 400. The strongest tier is reserved for chestpieces. Many armor pieces had their max integrity values adjusted.
 - Iron plate as well as normal (steel) plate armor had their armor values slightly shifted down by 5. Halberd stab's AP went from 40 to 45. This should finally allow a 11 strenght halberd stab to pierce just a tiny bit through normal plate armor. Good plate retains its old values since it's only present on snowflake pieces of equipment anyway.
 - Masks (iron mask, steel mask) now both share 'bad' plate protection, the difference being in their maximum integrity and crit protection flags. Previously, masks were being used as strictly better visors, as they didn't cover as much of your vision, had a separate healthbar and extremely high protection values too.
- Removed the stab crit protection flag from iron chainmail variants. Historically, it was fairly easy to break apart with appropriate weaponry (which, if a stabbing weapon is getting through its protection, it is). This should also make steel chainmail more relevant than just 5 extra protection value.
- All shirts should now have bite crit protection flags. This is really just meant to avoid scenarios of spawning in and getting immediately rushed by a wolf who then lands a lucky heart crit.

### Weapons:
**Polearms:**
- The spear gains a quicker thrust with little AP, now aptly named a 'poke'. Its heavier thrust is still firmly 10 AP behind other polearms, given how common of a weapon it is. When one-handed, it cannot deliver its heavier thrust.
- Polearm chop gains some AP to be a little more competitive with the thrust. Better at breaking armor, still not better at penetrating it - plus higher misscost, swing delay, charge time.. You get the point.
- Halberds no longer magnetically attach to your back. Eagle's Beak loses the equipped slowdown (unique weapon, not as strong as a halberd).
- Billhook and Eagle's Beak both gain +3 damage over the spear. 28 damage, same as a longsword.
- Makes the 'polearm swing', exclusive to the quarterstaff line of weapons, actually usable. 2 range, same charge time as a thrust, base damage factor.
- On that note, the staffs had been knocked down to 15 damage wielded, plus minimal adjustments. This sets back the wielded wood staff damage to a respectable value (23), with the steel quarterstaff now resting at 25 damage wielded.
- The bardiche now has appropriate, two-handed axe damage.
- The scythe gets a minor buff to account for the sucky intent. Still nowhere near a proper weapon.

**Swords:**
- Gives the arming sword variants a pommel strike intent. Rejoice. Not a thing when wielded, no half-swording these.
- Sabres and curved swords in general now get their own snowflake cut and chop intents, with higher AP (23 and 18 respectively). This should let them fulfill their roles as specialized cutting blades without crumbling when faced with literally any clothing or leather.
- Sabres get their chopping intent back (as above). They are also swift weapons once more. The cutlass and the king's sword even have a pommel to strike with.
- Shalal saber loses its.. Two tile range, longsword stab? I think this was an oversight, it's still unique now as the only sabre that can be two-handed.
- Some minor changes to scimitars, falchions are now better at parrying to give them some niche.
- The 'estoc' (the iron rapier) is now called a florete. Estoc as a name should really just be reserved for a two-handed specialized thrusting sword, which we don't have right now.
- The rapier, previously 40 AP, now 25, goes to a healthier middle ground of 30 AP. Keep in mind it has no pommel strike, a bad cut and no chopping intent.
- Sword chopping no longer has a steep swingdelay, except in the case of greatswords. Chargetime is retained for longswords and greatswords.
- Greatswords redesigned as specialized chopping blades.
 - Minimum strenght required to hold in one hand was increased in order to give them a higher damage when held in a single hand. Mostly a gimmick, doesn't have balance implications.
 - When wielded, has access to the default sword chop (no chargetime or swingdelay, low AP, 1.1 damage), or a new greatsword chop (halberd chop charge time, lowish but existant AP, 1.4 damage). The later serves to cement the greatsword in its own niche now as the best weapon against unarmored or lightly armored targets. Against proper armor, thrusting polearms still outpace it due to a better crit flag (less armors protect against stab than they do against chop) and better AP.
 - Greatswords lose their stab intent as a tradeoff for this, as the whole thing is simply too heavy to be a practical thrusting weapon.
 - The zweihander is mostly spared from this madness.
 - Iron messers and iron swords no longer smelt into steel.

**Other weapons:**
- AP values on maces/blunt weapons shifted up. A heavy smash in particular should have no issue breaking bones through chainmail now, as God intended. The dwarven warhammer gets a little extra parry as a signature weapon.
- Minor adjustments to axes. The battleaxe and bearded (steel) axe now have access to a charged chop intent with higher AP. The poleaxe's charged 2-range chop now has an actual damfactor attached.
- Stabbing is now the first intent on daggers again. Very minor thing that I personally felt, often cutting on accident since the last changes to weapons. Steel dagger goes back to 14 damage.
- Shield bashing now has club AP attached.
- The generic 'pick' intent was being used on some weapons for.. Some reason. It was pretty abysmal, with a swingdelay of 3 and no AP attached. So I gave it some armor penetration. This isn't on any meta weapon and is absolutely impractical so I doubt it'll matter.

## Why It's Good For The Game

The reasoning for each change has been provided at lenght, but in general, this should help re-diversify weapons and make each of them more individually satisfying.
On the armor side of things, it'll make sub-area aiming more important as chestpieces retain the highest integrity values and thin sections such as the hands or neck are appropriately statted to be more fragile. Same rationale goes for masks, making helmet visors more relevant and desirable.

## Things To Do

- Gather feedback.

## Pre-Merge Checklist

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/swords.dm
+++ b/code/game/objects/items/rogueweapons/melee/swords.dm
@@ -278,6 +278,7 @@
 	force_wielded = DAMAGE_SWORD_WIELD-1
 	desc = "A simple iron sword with a tested edge, sharp and true."
 	icon_state = "isword"
+	smeltresult = /obj/item/ingot/iron
 	max_blade_int = 200
 	max_integrity = INTEGRITY_STRONG
 	wdefense = GOOD_PARRY
@@ -422,6 +423,7 @@
 	icon_state = "imesser"
 	possible_item_intents = list(/datum/intent/sword/cut, /datum/intent/axe/chop)
 	gripped_intents = list(/datum/intent/sword/cut, /datum/intent/axe/chop, /datum/intent/sword/thrust)
+	smeltresult = /obj/item/ingot/iron
 	wbalance = EASY_TO_DODGE
 	sellprice = 20
 


code/game/objects/items/rogueweapons/intents.dm
```diff
@@ -286,6 +286,7 @@
 	swingdelay = 0
 
 /datum/intent/pick
+	penfactor = 50 //This is used in some actual weapons, so giving it some AP is likely wise. Hefty swing delay means it is impractical in combat.
```
NPC1314, line 289: should be in the AP_defines for readability & tweaking

code/game/objects/items/rogueweapons/melee/swords.dm
```diff
@@ -423,9 +442,9 @@
 	minstr = 6
 	wbalance = VERY_HARD_TO_DODGE
 
-/obj/item/rogueweapon/sword/rapier/ironestoc
-	name = "estoc"
-	desc = "A precise iron estoc, favored by the skilled duelists of Valoria."
+/obj/item/rogueweapon/sword/rapier/iron
```
NPC1314, line 445: why changing the name though? Estoc is a real thing after all

anything86: More of a nitpick than anything else, estocs are two-handed weapons specialized to puncture chain and plate. Basically a bigger, thicker, more rigid rapier. I think florete gets across better that this is not an upgrade to the rapier in any way, and frees up the name for a potential estoc to be implemented.

code/game/objects/items/rogueweapons/intents.dm
anything86: I will add this and the sabre intents to the AP defines on second thought.

NPC1314 (approved):

PR #1156, opened 2024-11-24
Assorted Dun Manor Map Tweaks

bowlofcereal:
## About The Pull Request

Changes a few rooms around in places like the Keep, adds the player-book shelves around some library areas (a few of the unique ones in certain areas, like the eoratica in the bathhouse), gives the Archivist room a bit of a nicer touch, changes the alchemical lab in the basement to a printing press room (Dun Manor doesnt have one from what ive seen), moves all alchemy ingredients to the upstairs Court Mage lab (which has been resorted to make alchemy more a prominent activity), and a few other minor touches to things I generally did not like (changing the randomized water barrels to just be normal water barrels in places like the tower, the keep serving quarters, etc.). A few other things that likely won't be noticed unless the keen eye pokes around, as they don't really change much of the game at all.

tl;dr changing some barrels around, moving the alchemical ingredients in the keep basement to the magi, turning it into a printing press room, making the archivist's room just that bit nicer, adding the randomized bookshelves for player books, etc etc.

## Why It's Good For The Game

Adds a few features that were not present on current maps. I'll be doing this for Raytown soon as well. Most of this is stuff the average player is not going to notice, so its mostly QOL (minus the printing press room). Should work completely fine as a result, it's not much of a change to the status quo in Dun Manor. I'll only be doing larger changes to Raytown since its not directly owned by anyone and I've declared my intent to do mass changes to it whenever necessary.

## Pre-Merge Checklist

### DIFF
--- a/_maps/map_files/roguetown/roguetown.dmm
+++ b/_maps/map_files/roguetown/roguetown.dmm
@@ -170,7 +170,7 @@
 /turf/open/floor/carpet/royalblack,
 /area/rogue/indoors/town/church)
 "ajg" = (
-/obj/item/roguebin/water,
+/obj/item/roguebin/water/gross,
 /turf/open/floor/rogue/tile,
 /area/rogue/indoors/town/manor)
 "ajp" = (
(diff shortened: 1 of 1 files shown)


lavillastrangiato:
Erm acktually

This might be out of the scope of this PR, but you should make the printing press and the library in general *outside* of the keep, so players can generally access it and upload their submissions without needing to play archivist. A huge library would be kind of banging tbh, and you could put the archivist's room and an alchemy lab in the library as well. (Please don't get rid of archivist alchemy lab access... cries)

bowlofcereal:
> Erm acktually
> 
> This might be out of the scope of this PR, but you should make the printing press and the library in general _outside_ of the keep, so players can generally access it and upload their submissions without needing to play archivist. A huge library would be kind of banging tbh, and you could put the archivist's room and an alchemy lab in the library as well. (Please don't get rid of archivist alchemy lab access... cries)

I can probably do that later but for now this is the current scope, yeah. I'll get around to it when I have another spark of inspiration, since it's not a really bad idea, just didn't like how the alchemy lab downstairs somehow had a better setup than the mage's own setup.. so i'm going to merge it for the sake of merging it to make space for a printing press. I'll probably find a spot to make a more public library for public use. The Archivist is always welcomed to come visit the Magician though for some alchemy lessons..

NPC1314 (approved):

PR #1205, opened 2024-12-02
The Traitening Pt. 1

anything86:
## About The Pull Request

Ports a few traits and adds a few others and gives them to an assortment of classes.

These are, alongside the classes that receive them:
- **Beautiful:** Grants a positive moodlet to whoever examines you and sees your face. Guaranteed for courtesan consort and prince/princess, high chance on highborn courtesan, middling chances on a few other roles such as bathmaiden and prisoner (which, ironically, has their face hidden...).
- **Swift Runner:** Removes collisions when running into walls and other structures and makes you not trip when running over rocks. Basically makes it harder to screw up a getaway, and makes running a safer option. Given to rogue, assassin, ranger and amazon (the last two use it for running amidst trees).
- **Good Runner:** Halved stamina cost while running. Currently only given to amazon.
- **Kick-Up:** Standing up from a prone position is 4x faster, though this reduction does not account for the time in which you cannot get up following a knockdown (basically, it only reduces the do_after). Given to rogue, qatil, swashbuckler and rogue adept.
- **Leaper:** Halves jump cooldown, allows jumping while laying down on the ground. Currently only given to Jester, mostly a meme.
- **Longstrider:** Negates slowdown on muddy tiles. Not a new addition, but originally only accessible via a spell. Now it's given to a multitude of roles that spend a long time out in the wilderness, such as hunter, briar, ranger and sentinel. Also given to mayor as a small buff, and because they are supposed to be policing the outskirts.
- **Martial Eye:** Allows you to gauge an opponent's speed relative to yours, as well as their level of exhaustion. Given to more brainy or experienced fighters, such as Captain, Veteran, Hand, Inquisitor and Treasure/Witch Hunter.
- **Blindfight:** Allows parrying outside one's cone of vision. Currently given to no one. Good candidate for a potion, _maybe_ should be given to Vampire Lord.
- **Iron Comfort:** Removes the slowdown associated with wearing armor. Given to Longbeard and Dwarf Underdark mercenary.
- **Shield Expert:** Guarantees blocking ranged attacks so long as a shield is in one hand. Still doesn't block bullets, and shields may break as usual. Given to Royal Guard and Hoplite.
- Warrior gets combat ready (no stamina drain from combat mode) as a provisory buff.

Barbarian rage is reworked. Now grants 45 seconds of faster attack speed (75% of the click CD, compared to moondust's 50%), +1 strenght and +4 endurance, as well as refreshing the user's fatigue (green bar) on activation and refreshing a little bit of stamina (blue bar) as well. As a tradeoff, barbarians cannot parry at all while raging, and the resulting debuff after it wears off is more impactful at -3 speed, and only goes away on sleeping.

This PR also removes the hardstun from getting thrown into a wall, which before allowed for cheesy strategies such as grabbing and repeatedly throwing into a wall. Could be brought back with a shorter stun as a monk specific trait, we'll see.

Also rogues to back to skilled daggers instead of skilled clubs since mace-wielding on a class with a strenght malus was non-sense, and weapon skill is hardly a priority if the intent is to be knocking out from behind.

## Why It's Good For The Game

This PR aims to increase role variety and expand the ways in which different classes can be balanced between eachother. Amazon, for example, was almost strictly inferior to Barbarian prior to this. Now with a few unique traits, there's some reason to pick it over Barbarian.

## To-Do
- Gather feedback, maybe shift some of the traits around.
- Martial Artist trait for Monk to allow punches to damage armor among other things.

## Pre-Merge Checklist


ThePainkiller:
- Barbarian should also get combat ready too tbh
- Sentinel should get Longstrider as well
- Leaper needs a paralyze or unconscious check as well so that jesters can't just backflip WHILE STUNNED

PR #1232, opened 2024-12-05
Hamlet polishing

NPC1314:
## About The Pull Request

Couple stairs fixed/improved
The alt dwarf outpost map wasnt activated, fixed
Lords key moved from being premapped to the lords inventory
since having it mapped when lord req gone means its just stolen by rando squires and whatnot every round, not the intended use.

## Pre-Merge Checklist

### DIFF
--- a/code/__HELPERS/icons.dm
+++ b/code/__HELPERS/icons.dm
@@ -1106,8 +1106,10 @@ GLOBAL_LIST_INIT(freon_color_matrix, list("#2E5E69", "#60A2A8", "#A1AFB1", rgb(0
 		if (isfile(thing)) //special snowflake
 			var/name = sanitize_filename("[generate_asset_name(thing)].png")
 			register_asset(name, thing)
-			for (var/thing2 in targets)
-				send_asset_async(thing2, key)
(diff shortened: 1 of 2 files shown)


code/modules/jobs/job_types/roguetown/nobility/lord.dm
```diff
@@ -42,6 +42,7 @@
 	backr = /obj/item/storage/backpack/rogue/satchel
 	belt = /obj/item/storage/belt/rogue/leather/plaquegold
 	backpack_contents = list(/obj/item/rogueweapon/knife/dagger/steel/special = 1)
+	neck = /obj/item/roguekey/lord
```
BadAtThisGame302, line 45: On every other map the key is mapped in, so this will cause duplications if the map is switched.

NPC1314: trivial to delete on the other maps, hold my beer.

NPC1314:
another funny thing is how lords key dont work on the goldface to prevent court just making merchant obsolete then hand and various others get merchant keys making that pointless.

PR #1240, opened 2024-12-10
Job display groupings

NPC1314:
## About The Pull Request

Parry adjustment works as intended, instead focus on grouping the jobs as
Court
Temple
Towners
Peasants
Outsiders
(apprentices)

because it makes a lot more sense and can be tied to mechancis in a smoother way than having mercenaries and butlers clumped with peasants. Also having a noble category with like 10 roles for a small sized town/court play is silly, there literally only needs and should be one noble family, the ruler. The new groupings also makes it easier to control antag selection etc etc.

## Pre-Merge Checklist


NameChangeStonekeep:
The niteman/nitematron has a noble status so they might fit the court more of their noble status

PR #1270, opened 2025-01-29
Spin emote actually spins

LesPirouettes:
## About The Pull Request

Source: https://github.com/Vanderlin-Tales-Of-Wine/Vanderlin/pull/529

Spin emote actually spins. Makes you vomit if you spin too much.

## Why It's Good For The Game

honestly im not sure.

## Pre-Merge Checklist


ThePainkiller:
In my experience on TG, this emote has been historically been used as a /wave to have two players communicate silently with one another, I could write it off as simply being flavor for dancing, but I hope it doesn't get missused just for the sake of it. The vomit is a good deterrent (even tg has this change iirc), but still, a matter of hoping our players aren't like that :)

ThePainkiller (approved):
For clarification: You only vomit if you were dizzy before, which can happen at a 20% with each spin, and it has an internal cooldown of 5 seconds to prevent spam. I like it. Good to go.

PR #1288, opened 2025-02-04
Bringing back Nitematron

Aegledis:
## About The Pull Request
Allowing female characters to play Nitematron again.
## Why It's Good For The Game
Player freedom + Guayo approved :3
## Pre-Merge Checklist

### DIFF
--- a/code/__DEFINES/jobs.dm
+++ b/code/__DEFINES/jobs.dm
@@ -134,7 +134,7 @@
 #define INNKEEP		(1<<5)
 #define COOK		(1<<6)
 #define TAILOR		(1<<7)
-#define NIGHTMAN 	(1<<9)
+#define NIGHTMAN 	(1<<8)
 #define NITEMAIDEN	(1<<9)
 #define ALCHEMIST	(1<<10)
 #define SCRIBE		(1<<11)


sanshoom:
does whore+ really need to be re-added, cause its all they ever did anyway.

code/modules/jobs/job_types/roguetown/nobility/nightman.dm
```diff
@@ -1,23 +1,25 @@
...
 	total_positions = 1
 	spawn_positions = 1
+	f_title = "Nitematron"
 
 	allowed_races = list(
 		"Humen",
 		"Dwarf",
 		"Elf",
 		"Half-Elf",
+		"Tiefling",
```
LesPirouettes, line 15: I don't mind female nitemaidens but tiefling/delf nobility feels like a change that should be documented.

DAKKA-WAAAGH:
> does whore+ really need to be re-added, cause its all they ever did anyway.

Nightman players have always been manwhores, it's not a shocker that the opposite gender would do the same

My opinion? A better discussion would be whether or not the role should still be part of the Nobility, when they never appear in court for anything, whereas most other nobles normally do. They'd rather sit in the bathhouse all round and maybe go out to the Merchant to buy expensive stuff with the ERP money. xd

Hot take; neither should the Merchant (a guild member, not a noble) or the Archivist (aka Book Servant).

BadAtThisGame302:
> does whore+ really need to be re-added, cause its all they ever did anyway.

So we should remove Nitemaster as a whole, then? Nitemasters have been whores+, we should remove males from the role.

code/modules/jobs/job_types/roguetown/nobility/nightman.dm
Aegledis: Just brought it back like it used to be but yea

PR #1291, opened 2025-02-04
Humen ancestries Revert

Aegledis:
## About The Pull Request

Humen ancestries reverted back to what they used to be + Metalklang dwarf ancestry for grenzie merc.
## Why It's Good For The Game

Guayo asked 👍 

## Pre-Merge Checklist


ThePainkiller:
Grenzel accent removal I feel is unnecessary. It should be assigned if you play a Grenzelhoft merc, and we could use it for future character customization revamps tbh.

sanshoom:
Why? Don’t these just add accents? Arguably more QOL than anything else. It shouldn’t get removed.

Aegledis:
> Why? Don’t these just add accents? Arguably more QOL than anything else. It shouldn’t get removed.

It links skintones with accents and makes makes people from a certain region have a specific skintone (example: everyone from Zybantu is forced to have darker skin)

PR #1292, opened 2025-02-05
Nukes pq

DovaKitty:
## About The Pull Request
Reduces the PQ requirements for towner classes down to 0 or lower, except for Veteran, which has been raised to 10.

## Why It's Good For The Game
It's time to admit that locking classes to new players on a whitelisted server was a mistake. It is especially terrible on low population. If it's to be re-evaluated, that should be done when pop increases to acceptable levels once more.

## Pre-Merge Checklist

### DIFF
--- a/code/modules/jobs/job_types/roguetown/temple/priest.dm
+++ b/code/modules/jobs/job_types/roguetown/temple/priest.dm
@@ -23,7 +23,7 @@
 
 	display_order = JDO_PRIEST
 	give_bank_account = 115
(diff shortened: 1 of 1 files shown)


ThePainkiller:
Is Blacksmith being a role that can be played even while softbanned for shittery an intentional change to try and make the job more played? I just hope it doesn't attract grifters to it. If anything, Beggar should be one of, if not the only, default negative PQ role.

PR #1297, opened 2025-02-10
New Flaw - Prosthetic Legs

WWWWW-Git:
## About The Pull Request
This PR adds a new flaw as well as expanding upon prosthetic legs. Every race now has a prosthetic leg sprite, essentially a pegleg. The 'Missing Leg' flaw can now be selected on character creation, similar to the 'Missing Arm' flaw. In place of a missing leg is a prosthetic one, the downside being that you cannot run if you have a prosthetic leg. Likewise if a player with a leg loses one and then has a prosthetic leg attached they will lose the ability to run.

pegleg

## Why It's Good For The Game

Despite prosthetic legs being a thing, they weren't ever really expanded upon or used, and no race had corresponding prosthetic leg sprites. I believe prosthetic legs allows for more RP opportunities, both as the result of something like a dungeon dive gone wrong, to someone just starting off with the flaw.

## Pre-Merge Checklist


sanshoom:
Thar’ be treasure…

